PR #3, opened 2016-09-13
Added IJ-dependent recentering of WFs in the following way:

paulatz:
1. three subroutines are added to hamiltonian.F90, the most important one is wz_translate_dist, given WF i at 0 and WF j at R it computes the basis vector of the supercell lattice that minimizes the 0+ri -> R+rj distance. Note that this depend on i,j and R and it is equivalent to putting WF R+rj inside the super-Wigner-Seitz centered in i and it is NOT equivalente to put both 0i and Rj inside the super-WS centered in 0.
2. if some shifts are equivalent, i.e. if WF Rj falls on the edge of the super-WS, all the possibilities are stored.
3. The shift vectors are stored in wdist_shiftj_wsi(3,ndegenx,num_wann,num_wann,nrpts); the number of equivalent shifts is stored in wdist_ndeg(num_wann,num_wann,nrpts)
4. in plot.F90 the WFs are shifted accordingly to get the best Fourier interpolation

NOTE: nothing is done in postw90, as these vectors need to be either stored or recomputed. It is also possible to change the real space hamiltonian to include the shift (in this way the R->G interpolation becomes trivially a sum) but this changes the number of R vectors, hence I have not done it here (I have done it for phonons in a special version of q2r)!

### DIFF
--- a/src/hamiltonian.F90
+++ b/src/hamiltonian.F90
@@ -27,6 +27,14 @@ module w90_hamiltonian
   !                   irvec(1:3,irpt) in the basis of the lattice vectors
   !
   integer,          public, save, allocatable :: irvec(:,:)
+  integer,          public, save, allocatable :: shift_vec(:,:)
+  !
+  ! [lp] The number of unit cells to shift WF j to put its centre inside the Wigner-Seitz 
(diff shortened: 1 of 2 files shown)


paulatz:
putting request on hold as a bit of refactoring is required for postw90

PR #54, opened 2016-09-16
Update geninterp test case for more reproducible band derivatives

jryates:


giovannipizzi:
I will close this as it was fixed in parallel in #52

PR #59, opened 2016-09-16
Debug for the parallel test !

sponce24:


giovannipizzi:
In the taxi I imagined that `export TMP` might have been the problem :-)

PR #76, opened 2016-11-25
Parallelization of wannier90.x (disentanglement and wannierise)

guillaumegeranton:


jryates:
Looks ok to me. Two issues that can be fixed post merge:
 Need to fix the Makefile
 Need to read the parameters on the root node and bcast (as is done in postw90)

PR #86, opened 2016-12-26
Added comments for FORD documentation

jryates:


jryates:
For a preview of the current code documentation see http://www.wannier.org/ford
I've done most of the code except - transport (due to ignorance), site symmetry (need to read the original papers, or ask the authors to do it), some of parameters (probably needs a joint effort).

PR #90, opened 2016-12-30
Fix use of uninitialised variable with nnkp parameter block

jryates:


giovannipizzi:
Ok, this should also fix the compilation configuration that was failing on the test farm

giovannipizzi:
Ops... I closed it instead of merging it... I wait for the tests and I merge it.

PR #96, opened 2017-01-25
G0W0 interface

AntimoMarrazzo:
G0W0 interface for Wannier90 (tested with QE and Yambo), with *no changes* to the Wannier90 main source code. New features:
-Two new utilities: gw2wannier90.py and k_mapper.py.
-New example 23 about G0W0 bands interpolation of silicon with QE and Yambo.
-Tutorial and user guide updated.
Part of the interface Yambo-Wannier90 which was very specific to Yambo has been implemented in the Yambo code itself (actually in Ypp, Yambo's post processing code) and it has just been added to latest release of Yambo.  The utilities are quite general, especially gw2wannier90.py can probably be used with any G0W0 code.


utility/gw2wannier90.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python2
+#
+# gw2wannier90 interface
```
giovannipizzi: Can you add this text in the comments, for the license (if it's ok to license under GPL)?

```
# This file is distributed as part of the Wannier90 code and
# under the terms of the GNU General Public License. See the
# file `LICENSE' in the root directory of the Wannier90
# distribution, or http://www.gnu.org/copyleft/gpl.txt
#
# The webpage of the Wannier90 code is www.wannier.org
#  
# The Wannier90 code is hosted on GitHub:
#
# https://github.com/wannier-developers/wannier90
```

utility/k_mapper.py
```diff
@@ -0,0 +1,74 @@
+#! /usr/bin/env python
+#
```
giovannipizzi: Can you add also here the license (same text as above?)

CHANGE.log
```diff
@@ -4,6 +4,9 @@
 
 	The Maximally-Localised Generalised Wannier Functions Code
 
+v2.x.y (25th January 2017)
```
giovannipizzi: Also, can you add a line to the files `examples/README`?

PR #97, opened 2017-01-25
Bug fix for writing uHu and uIu matrices

huachenUTexas:
Bug fix for writing uHu and uIu matrices in pw2wannier90. The original code does not take into account the case of nonzero exclude_bands.


jryates:
Pushing this into a separate branch where we can sort out various pw2wannier90 issues

PR #105, opened 2017-03-08
Berry optimization

ponweist:

PR #108, opened 2017-03-10
Combined reqeust for #105, #106 and #107

ponweist:
In case you plan to accept all three pull requests #105, #106 and #107, I've created this single combined request for convenience (I noticed that accepting them one after the other will result in minor merge conflicts).


hjunlee:
There seems to be a bug in the subroutine get_gauge_overlap_matrix in get_oper.F90.
Since the first dimension of eigval is num_bands, 
in the argument of call utility_zgemmm
eigval(:,ik_a) should be eigval(wm_a:wm_a+ns_a-1,ik_a).

H. Lee

Issue #101, opened 2017-02-16
Thresholds in ws_distance.F90


greschd:
Email conversation: 
On Thursday, February 16, 2017 1:34:42 PM CET Dominik Gresch wrote:
> Thank you very much for your reply. Indeed, increasing the tolerance (in
> my case, to about 7.d-4) fixes this issue. I changed the code such that
> the tolerance can be given as an input parameter, see here:
> https://github.com/greschd/wannier90/tree/irdist_ws_tol
>
> First of all, do you think this is useful / necessary in general? If
> yes, are there other places in ws_distance.F90 where the tolerance
> should be changed accordingly? From glancing over the code I would guess
> maybe eps = 10 * irdist_ws_tol would make sense. Of course we could also
> let the user set eps and use irdist_ws_tol = eps / 10 . Please let me
> know what you think about this.

@paulatz wrote:

All the thresholds are inside ws_distance, there should be nothing outside of 
this file, which is good.

You are right that the various thresholds should be consistent, although I'm 
not sure it is trivial to do because some are measured in direct lattice 
units, other in fractional units.

In particular :
1. in R_ws_sc_equiv, line 235, we are accepting a **relative** error of 10^-5 in 
cartesian coordinates (coordinate system is important for anisotropic 
lattices)

2. in ws_translated_dist, line 157 we require an **absolute** precision of 10^-6 
in fractional coordinates.

If I remember correctly, I put this second check in order to spot bugs in the 
code, but now I think the check in ws_translate_dist could just be removed.

I think it could be a good idea to have the threshold value in R_ws_sc_equiv 
be an input parameter, and maybe default to something larger than 1.d-5

Also, I'm not sure that using a relative threshold, instead of an absolute 
value is justified, because it would require the centres of the WFs to be less 
accurately equivalent when using finer grids of k-points.

greschd:
I would suggest the following: 
* Change the first test from relative to absolute, and making it configurable with a sensible default value. I think this is what the end user should care about -- how similar should the distances be to be considered equivalent.
* Keep the second test, but change the value such that it's really just a sanity check and should never fail, even considering numerical inaccuracies (something like 0.05). As far as I can see, this is just to test that rounding to the nearest integer doesn't change the value too much.

greschd:
Of course we also need a good name for the input flag. Some candidates:
- ``ws_dist_tol``
- ``ws_distance_tol``
- ``ws_dist_equiv``
- ``ws_distance_equiv``
[...]

PR #109, opened 2017-03-11
Ws distance tol

greschd:
This PR adresses issue #101 . The changes are as follows:

* The check in ws_distance.F90 (line 156) is made more lenient, to avoid false errors
* The check for positions of equivalent distance (line 236) has two changes:
  * The absolute difference in distance is calculated instead of the relative one
  * The tolerance can now be changed by the user (``ws_distance_tol`` parameter)

The reason I'm marking this as work in progress is that the change in the second check breaks the test ``si_geninterp_wsdistance``. I have investigated this, and it seems to be because the Wannier positions are very close, but not exactly equal, in the tested example. Due to the change from relative to absolute distance there's no value for ``ws_distance_tol`` which exactly keeps the previous behavior. @paulatz , what is your take on this? Can we safely replace the test data?


paulatz:
Hello, I got a bit distracted away from this issue. I think what Dominik proposes is correct, the consistency check should not be necessary and can cause problems. I think we can replace the test data with no problem. However, on my computer using the Dominik version (but maybe I'm not cloning the correct git branch) or the upstream one gives the same result for the test. I guess we may be close to the tolerance, maybe it would be better to increase the convergence threshold of the minimization, to avoid having to avoid having spurious test errors?

paulatz:
Ok, I managed to find the new version of the code an I tested it extensively. It is indeed impossible to reproduce the same results of the test because of the change of the threshold from relative to absolute, bit it does not make a real difference. 

The results are actually almost identical when using ws_distance_tol=5.d-5. It does not really make any difference if two WF which are almost on the boundary of each other's WS cell are put inside or outside, or on the border.

Personally, I approve this new version from Dominik and updating of the test.

greschd:
Thanks for testing Lorenzo. I updated the PR with the new test results.

greschd:
Alright, I added the section describing the input parameter. For the unit: the cartesian coordinates are always in Angstrom internally, correct?
Yes, this closes #101.

As a side note, the "plot parameters" table is now too long to fit on its page. I'm guessing we'll fix that before the next release (since there might be more parameters coming), right?

PR #120, opened 2017-03-29
Update postw90_common.F90

hjunlee:
Currently, Wannier90 spends the useless memory in some parts, especially in the parallel run.
For this reason, when considering systems with large number of k points and bands, it might lead to memory error.

In addition to this, there are some points in which memory is used uselessly and they are also related to the parallelisation of both core parts (wannierization and disentanglement) and post-processing parts. 
Although I have my working parallel version of Wannier90, I will make an effort to apply, if any, the difference to the already-committed version.

H. Lee

### DIFF
--- a/src/postw90/postw90_common.F90
+++ b/src/postw90/postw90_common.F90
@@ -437,7 +437,8 @@ module w90_postw90_common
                                io_date,io_time,io_stopwatch
     use w90_parameters, only : num_wann,num_kpts,num_bands,have_disentangled,&
                                u_matrix_opt,u_matrix,m_matrix,&
-                               ndimwin,lwindow,nntot
+                               ndimwin,lwindow,nntot,&
+                               num_valence_bands,scissors_shift
 
     implicit none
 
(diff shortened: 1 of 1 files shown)


ponweist:
A viable way to reduce the memory requirements may be using MPI shared memory windows for storing read-only matrices only once per computation node (not once per process). 

You may have a look at the example [hpc_kernel_samples/parallel_io/read2shmem](https://gitlab.com/PRACE-4IP/CodeVault/tree/master/hpc_kernel_samples/parallel_io/read2shmem) from the [PRACE CodeVault](http://www.prace-ri.eu/prace-codevault/), which explains and demonstrates the usage of MPI shared memory windows.

hjunlee:
Thank you very much for the info.
Currently, I solve the memory issue in core parts with parallelization enabled by allocating some large matrices (for instance, m_matrix, m_matrix_orig, etc.) only in a root node and using mpi_gatherv, mpi_scatterv, etc.
Regarding the above patch, currently there is no need to maintain the above matrices in postw90 calculations.

PR #124, opened 2017-03-30
76 develop fix mpi

jryates:
Modifications by guillaumegeranton to use MPI in the core wannierise and disentangle routines. Tested in serial and parallel and a few bug fixes added.


jryates:
Superseded by #125

PR #126, opened 2017-03-31
Update get_oper.F90

hjunlee:
There seems to be a bug in the subroutine get_gauge_overlap_matrix in get_oper.F90.
Since the first dimension of eigval is num_bands,
in the argument of call utility_zgemmm
eigval(:,ik_a) should be eigval(wm_a:wm_a+ns_a-1,ik_a).

H. Lee


jryates:
Do you have a small test case that highlights this problem? It would be great to add this to the test suite so we know the bug stays fixed.

hjunlee:
I don't have small test case, but it leads to error whenever get_BB_R is called, for example, plot_morb is true, and the lower bound of the outer energy windows rather than 1 is set.

ponweist:
This seems very plausible to me. I think I introduced the bug in ponweist/Wannier90-PRACE@804c8f9deb1e0f0d264a0816937cdc2a815142fa - thanks for fixing!

ponweist:
I'd recommend to accept this bug fix and merge it to the 108_POSTW90OPT_FIX branch.

PR #151, opened 2017-12-05
Fix 133 restart

### LINKED ISSUE #133: Problem in restart after merging the parallel code
From Arash:
With the benzene example, if I run a wannierisation, and then do a second run setting "restart = wannierise” (or indeed plot) I get an error 
```
forrtl: severe (24): end-of-file during read, unit 11, file /export111/work/aam/w90-examples/example12-val/benzene.chk
```

Or:
example01, example02 etc: instead of the end-of-file error on reading .chk files, I sometimes get the following error on the first cycle of the restart:
```
Cycle:      1
 wann_main: ZHEEV in internal_new_u_and_m failed, info=            3
            trying Schur decomposition instead
 wann_main: SCHUR failed, info=            4
 Exiting.......
 wann_main: problem computing schur form 1
```

giovannipizzi:
Adding a test to prove that #133 was fixed.
Closes #133


jryates:
Fine to me (just adds test)

jryates:
All fine! (press the correct button this time!)

PR #154, opened 2017-12-20
fix python script generated by kslice

qiaojunfeng:
Hi, when I run examples17 in the tutorial, I came up with a python error:

```
Traceback (most recent call last):
  File "Fe-kslice-fermi_lines.py", line 36, in <module>
    bbands=bands.reshape((dimy,dimx,numbands))
TypeError: 'float' object cannot be interpreted as an integer
```

The python script was generated by kslice calculation of postw90.x.
I'm using Anaconda 4.3.30 on openSUSE Tumbleweed.

I found simply add an int() conversion could fix the bug. 
Thanks for the wannier90, it's wonderful!

### DIFF
--- a/CHANGE.log
+++ b/CHANGE.log
@@ -4,8 +4,8 @@
 
 	The Maximally-Localised Generalised Wannier Functions Code
 
-* bug fix in kslice.F90 & example17 (missing type conversions when plotting Fermi 
-  lines in the python scripts)
+* bug fix in kslice.F90 & example17 (change to floor division in the python 
+  scripts for plotting Fermi lines), minor bug fix in example18 & tutorial
 
 v2.x.y (DATE)
 * G0W0 interface implemented (A. Marrazzo (EPFL, CH) and S. Tsirkin (DIPC, Spain)).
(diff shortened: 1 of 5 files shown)


greschd:
Hi @qiaojunfeng,

This issue is due to division changing from integer division to "true" division from Python 2 to 3. While casting the result to ``int`` also works, the correct way of doing this would be to use the integer division operator ``//``. For Python 2 compatibility, you need to add ``from __future__ import division`` to the top of the file.

So it would be 
```
from __future__ import division

...

num_bands = bands_up.size // num_pt
```

qiaojunfeng:
Hi @greschd ,

You are right, it's better to use floor division `//` than int casting. I have changed it to `//`. Forgive my dirty hack.

Moreover, I found another two minor bugs,
1. in `example18/Fe.win` line 37, `kslice_kmesh` should be `kslice_2dkmesh`
2. in Tutorial.pdf of example18, page 33, `gnuplot> plot ‘Fe-kubo-A_xy.dat’ u 1:2 w l` should be `gnuplot> plot ‘Fe-kubo_A_xy.dat’ u 1:2 w l`, the filename should be `kubo_A` instead of `kubo-A`

jryates:
Thanks for fixing this!

PR #155, opened 2017-12-21
plot spinor-valued WFs

hjunlee:
Dear developers:

These files are for plotting spinor-valued Wannier functions.
I used them with QE v6.x and VASP v5.x and included pw2wannier90.f90 for QE v6.2.1 (I think that it can be used with QE v6.2 since there is no change in pw2wannier90.f90 between QE v6.2 and v6.2.1).
UNK files for spinor Bloch functions are formatted as follows:
format ('UNK',i5.5,'.','NC')

With these simple extension, we can plot spinor-valued WFs in three ways: total norm ("total"), norm of up-spinor WF ("up"), or norm of down-spinor WF ("down").
They are controlled by the new parameter (wannier_plot_spinor_mode); 
for example, wannier_plot_spinor_mode = up
In default, total is chosen.

Sincerely,

H. Lee


jryates:
Thanks! This addresses a long standing issue. I'm going to merge it into a branch, so that I can fix a compile issue, update the manual, and run some tests

jryates:
After running some tests it occurred to me that in the case of 'up' or 'down' it is better to write out a signed norm i.e. |phi|*sign(Re{phi}) where phi is either the up or down part of the spinor. This has the advantage that for, say, ferromagnetic iron, the WF plotted are visually very similar to the collinear ones.
 I thought about having a keyword to control this behavior (signed vs not signed) but decided against this as it is trivial to ignore the sign if you don't want it.
 i need to finish the documentation before making a pull request into the main branch, but any thoughts or suggestions welcome.

mostofi:
As the choice is essentially arbitrary, it might be good to have it as an
option, as different users may want different behaviour?

hjunlee:
Dear developers:

I have no objection, but I think that you had better add some comment about the possible issue in the manual.

>This has the advantage that for, say, ferromagnetic iron, the WF plotted are visually very similar to the collinear ones.

I am afraid that this similarity might be specific to this case; I am not sure whether this similarity can be achieved if we wouldn't start from the good initial choice of projections and the calculation of WFs wouldn't converge well.

I think that in general we need an additional step like that described in your paper (https://journals.aps.org/prb/abstract/10.1103/PhysRevB.74.195118) if we want to give the phase of spinor-valued WFs some physical meaning.

Also, I think that for materials with non-collinear magnetism, it is nontrivial to give the phase of spinor-valued WFs physical meaning.

Sincerely,

Hyungjun Lee

PR #160, opened 2018-01-03
Gyrotropic module

stepan-tsirkin:
The present pull request adds a new module, called "gyrotropic", which was used for the calculations, presented in the article: 
S. S. Tsirkin, P. Aguado Puente and I. Souza, arXiv:1710.03204 
Changes in the code are 
     New file gyrotropic.F90
     corresponding changes in postw90.F90, postw90_common.F90, parameters.F9
    added optional parameter "occ" in subrotines:
        berry_get_imf_klist
        berry_get_imfgh_klist
        wham_get_JJp_JJm_list
        wham_get_eig_UU_HH_JJlist
        wham_get_occ_mat_list (also reordered the parameters, so "eig" is now also opional)

  in io.F90 changed 'maxlen' to  255 

   added new functions:
       utility_det3
       spin_get_S

example24

Corresponding contribution to the User Guide and Tutorial


jryates:
Stepan - many thanks for this. Looks like there are a few small fortran/makefile issues to sort out.
 It would be good to have one or more test cases. They don't need to be physical- just run very quickly. If you want to send me some inputs, I'm happy to put the infrastructure together. The only think to watch is that some of the files (uHu in particular)  are rather large - so at the moment all of the Berry related tests share the same inputs (Fe, of course).

stepan-tsirkin:
Hi, Jonathan. Sorry for not responding for two weeks. I was out of office these days. Now I have fixed the errors, the check is passed. Regarding the tests, Fe is not really suitable as a gyrotropic testcase, so the folder wannier90/test-suite/tests/testpostw90_te_gyrotropic/ contains  a test case for Te, which is a reduced version of example24.  (accuracy parameters are reduced to unphysically small values). It contatins all input files, wannier files, and the output, which I get. It amounts to 2MB, which I think is not really much. Could you make an automated test out of this, as I am not experienced in this.
Thanks for help, tell me if something is needed. I am happy to contribute to wannier90, and hope it will be useful for someone.

PR #162, opened 2018-01-16
Previous changes to plot.f90 from Arash 

VVitale:
Changes made to generate cube files for non orthogonal unit cells


jryates:
Valerio, Thanks for this. Could you also document the new keyword in the user guide please.
Jonathan

VVitale:
Hi Jonathan, sorry for the late reply. There is no new keyword to add, but I should probably change the sentence in the user guide where it says that only orthogonal lattice vectors are allowed when using a Gaussian cube format.

jryates:
It looks like there is a new input variable called wannier_plot_scale

VVitale:
Oops, sorry you are absolutely right!

PR #163, opened 2018-01-17
Code coverage and Gyrotropic Tests

jryates:
This adds hooks to use the codeconv code coverage tool. We also add various extra tests, including for the new gyrotropic module.


giovannipizzi:
Two quick comments
There is this message in travis:
```
--------------------------------------------------------------------------
[[9679,1],0]: A high-performance Open MPI point-to-point messaging module
was unable to find any relevant network interfaces:
Module: OpenFabrics (openib)
  Host: travis-job-wannier-develo-wannier90-341781176.travisci.net
Another transport will be used instead, although this may result in
lower performance.
--------------------------------------------------------------------------
```
(was there also before) - any idea on how to fix this (or if it is really a problem)?

And most importantly - the time for tests is now over 20 minutes, that makes the development process a bit slow especially since one has to run the tests, wait for them to complete before merging and going to the next PR. Do you think it would be possible to speed up the parallel tests somehow?

jryates:
Apparently this is openmpi looking for an Infiband interface, and falling back to ethernet when it doesn't find one. passing -mca btl ^openib to mpirun is supposed to solve it.

The parallel test are supposed to be quite fast. I'll have a look to see what the issue is.

PR #164, opened 2018-01-23
Bugfix for Symmetry-adapted WFs in compute_dmn within pw2wannier.f90 v6.1.2

VVitale:
1) bug in serial execution 
#if defined(__MPI) environment has been introduced between gather_grid calls and similar

2)  bug in read_sym mode
In compute_dmn subroutine, “nsym” variable is defined locally rather than globally which is fixed at scf/nscf calculation.

### DIFF
--- a/src/postw90/berry.F90
+++ b/src/postw90/berry.F90
@@ -1100,6 +1100,8 @@ module w90_berry
              joint_level_spacing=sqrt(dot_product(vdum(:),vdum(:)))*Delta_k
              eta_smr=min(joint_level_spacing*kubo_adpt_smr_fac,&
                   kubo_adpt_smr_max)
+          else
+             eta_smr=kubo_smr_fixed_en_width
           endif
           rfac1=(occ(m)-occ(n))*(eig(m)-eig(n))
           occ_prod=occ(n)*(1.0_dp-occ(m))


src/wannierise.F90
```diff
@@ -1878,14 +1885,19 @@ subroutine wann_domega(csheet,sheet,rave,cdodq)
...
                       ( crt(m,n) * ln_tmp_loc(n,nn,nkp_loc)  &
                      + conjg( crt(n,m) * ln_tmp_loc(m,nn,nkp_loc) ) ) &
                      * cmplx(0.0_dp,-0.5_dp,kind=dp)
-                cdodq_loc(m,n,nkp_loc) = cdodq_loc(m,n,nkp_loc) - wb(nn) &
+                cdodq_loc(m,n,nkp_loc) = cdodq_loc(m,n,nkp_loc) + wb(nn) * (lambdac - 1.0_dp) &
                      * ( crt(m,n) * rnkb_loc(n,nn,nkp_loc) + conjg(crt(n,m) &
                      * rnkb_loc(m,nn,nkp_loc)) ) * cmplx(0.0_dp,-0.5_dp,kind=dp)
+                cdodq_loc(m,n,nkp_loc) = cdodq_loc(m,n,nkp_loc) - lambdac * wb(nn) &
+                     * ( crt(m,n) * r0kb(nn,nkp_loc) + conjg(crt(n,m) &
+                     * r0kb(nn,nkp_loc)) ) * cmplx(0.0_dp,-0.5_dp,kind=dp)
```
jryates: I think there is something not right in calling r0kb with nkp_loc, either here - or earlier when the data is setup. It doesn't appear consistent in the two places.

VVitale: Hi, sorry about this. The commit for for the selectively localised WFs shouldn't be merged to the main repository yet. Not sure what is the best way to proceed now.

PR #167, opened 2018-02-14
SCDM method in Wannier90

VVitale:
Major commit for the SCDM method. Files modified are: parameters.F90 kmesh.F90 and pw2wannier90.F90, the latter being the one with the actual code for the projections.  
I have also added the necessary info in the documentation (user guide) on how to use the new projection method.


doc/user_guide/parameters.tex
```diff
@@ -611,7 +614,8 @@ \subsection{Shells}
 \section{Projection}
 
  The projections block defines a set of localised functions used to
- generate an initial guess for the unitary transformations. 
+ generate an initial guess for the unitary transformations. The projection block can be specified
+in conjunction with {\tt scdm\_proj=true} (see belove). This is only used to read the centers of the projections, which in some cases could help the optimisation if {\tt guiding\_centres=true} is added to the input file.
```
giovannipizzi: belove->below

doc/user_guide/parameters.tex
```diff
@@ -736,6 +740,31 @@ \section{Job Control}
 
 The default value of this parameter is $\verb#false#$.
 
+\subsection[scdm\_proj]{\tt logical :: scdm\_proj}
+If {\tt scdm\_proj=true} then the $A_{mn}^{(\mathbf{k})}$ matrices are generated with the SCDM-k method of Ref.~\cite{LinLin-ArXiv2017}. In this case, one also needs to specify the other three {\tt scdm} keywords. One then needs to run {\tt wannier90.x -pp seedname} to generate the {\tt seedname.nnkp} file, to be used by a first-principle code (at the moment only interface available is with the QuantumEspresso code). It is not possible to set {\tt scdm\_proj=true} and simultaneously specify a projection block.
```
giovannipizzi: Are mu and sigma really always needed to be specified, also in the isolated case?

src/parameters.F90
```diff
@@ -1898,6 +1904,39 @@ subroutine param_read ( )
...
+    scdm_mu = 0._dp
+    scdm_sigma = 0._dp
+    scdm_entanglement = 0
+    call param_get_keyword('scdm_proj',found,l_value=scdm_proj)
+    if(found .and. allocated(proj_site)) &
+        call io_error('Error: Can not specify projections and scdm_proj=true at the same time.')
+    if(found .and. scdm_proj .and. spinors) &
+        call io_error('Error: SCDM method is not compatible with spinors yet.')
+    if(found .and. scdm_proj .and. guiding_centres) &
+        call io_error('Error: guiding_centre is not compatible with the SCDM method yet.')
```
giovannipizzi: guiding_centres

src/parameters.F90
```diff
@@ -1898,6 +1904,39 @@ subroutine param_read ( )
...
+    if (found .and. (scdm_sigma < 0._dp)) & 
+       call io_error('Error: The parameter sigma in the SCDM method has to be positive.')
+    call param_get_keyword('scdm_entanglement',found,c_value=ctmp)
+    if (found) then
+       if(ctmp=='isolated') then
+         scdm_entanglement = 0
+       elseif(ctmp=='erfc') then
+         scdm_entanglement = 1
+       elseif(ctmp=='gaussian') then
+         scdm_entanglement = 1
```
giovannipizzi: should be 2!

doc/user_guide/parameters.tex
```diff
@@ -736,6 +740,31 @@ \section{Job Control}
...
+
+\subsection[scdm\_mu]{\tt real(kind=dp) :: scdm\_mu}
+The value of the $\mu$ parameter in the formulas above. It defines the characteristic energy for the occupation numbers matrix, in units of eV. If {\tt scdm\_entanglement=1}, it gives the mean value of the complementary error function. If {\tt scdm\_entanglement=2}, it gives the mean value of the gaussian instead. 
+
+The default value is {\tt fermi\_energy eV}
+
+\subsection[scdm\_sigma]{\tt real(kind=dp) :: scdm\_sigma}
+The value of the $\sigma$ parameter in the formulas for the occupation numbers matrix. It defines the spread of the occupation numbers matrix around $\mu$, and as such it cannot be negative. It must be given in units of eV. 
+
+The default value is {\tt 0.0 eV}
```
giovannipizzi: Not sure if setting the default to zero is a good idea, will give rise to division by zero. Better to give e.g. a default of 2eV, or 4eV? Or leave to zero, but then in parameters.F90 check if it is zero and complain that it has to be specified or > 0.

doc/user_guide/parameters.tex
```diff
@@ -736,6 +740,31 @@ \section{Job Control}
...
+\item[{\tt 1}] The occupation number matrix is given by: $$f(\epsilon_{n\mathbf{k}})=\frac{1}{2}\mathtt{ERFC}\left(\frac{\epsilon_{n\mathbf{k}} - \mu}{\sigma}\right) $$
+\item[{\tt 2}] The occupation number matrix is given by $$f(\epsilon_{n\mathbf{k}})=\mathtt{EXP}\left(-\frac{(\epsilon_{n\mathbf{k}}-\mu)^2}{\sigma^2}\right)$$
+\end{itemize}
+
+The default value is {\tt 0}.
+
+\subsection[scdm\_mu]{\tt real(kind=dp) :: scdm\_mu}
+The value of the $\mu$ parameter in the formulas above. It defines the characteristic energy for the occupation numbers matrix, in units of eV. If {\tt scdm\_entanglement=1}, it gives the mean value of the complementary error function. If {\tt scdm\_entanglement=2}, it gives the mean value of the gaussian instead. 
+
+The default value is {\tt fermi\_energy eV}
```
giovannipizzi: However, from the logic in parameters.F90, if the fermi_energy is not specified, the default remains at zero eV... maybe better to simplify the logic, and leave this value by default at 0 eV? If the user specifies by hand the fermi_energy flag, he can also set it twice here. Also, in our experience mu=fermi is not always the optimal choice.

doc/user_guide/parameters.tex
```diff
@@ -736,6 +740,31 @@ \section{Job Control}
...
 The default value of this parameter is $\verb#false#$.
 
+\subsection[scdm\_proj]{\tt logical :: scdm\_proj}
+If {\tt scdm\_proj=true} then the $A_{mn}^{(\mathbf{k})}$ matrices are generated with the SCDM-k method of Ref.~\cite{LinLin-ArXiv2017}. In this case, one also needs to specify the other three {\tt scdm} keywords. One then needs to run {\tt wannier90.x -pp seedname} to generate the {\tt seedname.nnkp} file, to be used by a first-principle code (at the moment only interface available is with the QuantumEspresso code). It is not possible to set {\tt scdm\_proj=true} and simultaneously specify a projection block.
+
+The default value is {\tt false}.
+
+\subsection[scdm\_entanglement]{\tt integer :: scdm\_entanglement}
+Select the functional form for the occupation number matrix $f(\epsilon_{n\mathbf{k}})$ for the SCDM-k method.
+Only three integer values are allowed:
```
giovannipizzi: This is not currently true, in the input the user specifies three strings - internally they are converted to an integer

pwscf/v6.2.1/pw2wannier90.f90
```diff
@@ -957,6 +969,45 @@ SUBROUTINE read_nnkp
...
+     IF (found) THEN
+        READ (iun_nnkp,*) scdm_proj_tmp
+        IF(scdm_proj_tmp==1) THEN
+           scdm_proj = .true. 
+           READ (iun_nnkp,*) n_wannier
+           READ (iun_nnkp,*) scdm_entanglement
+           READ (iun_nnkp,*) scdm_mu, scdm_sigma
+        ENDIF
+     ELSE
+        WRITE(stdout,'(//," ****** begin WARNING ****** ",/)')
```
giovannipizzi: Do we really want to issue a Warning? If the user does not want to use SCDM, then he does not need to upgrade to a recent version. If he wants to, then he will have to have already upgraded. 
This warning might "scare" users but I don't think it is needed (most probably they will read this and don't know what SCDM is).

doc/user_guide/parameters.tex
VVitale: done

doc/user_guide/parameters.tex
VVitale: No they don't you are right. I have modified the text according to the logic.

doc/user_guide/parameters.tex
VVitale: Now mu is 0.0 eV by default.

doc/user_guide/parameters.tex
VVitale: Now sigma is 1.0 eV by default. Then in parameters.F90 it checks if sigma is specified and it returns an error if it is not positive.

src/parameters.F90
VVitale: done

src/parameters.F90
VVitale: Oops! This one is more serious.

hjunlee:
Dear developers:

Although I tried to do the similar things in both USPP in QE and PAW in VASP, I am still wondering about the extension of SCDM to USPP and PAW where the norm of PW parts only is not equal to 1.

In the pw2wannier90, you represent only (soft) PW parts of wave functions on the real grid and you enforce normalization of psic_all by dividing norm_psi. As a consequence, since in general norm_psi is different between different band indices and thus, psic_all with different band indices are scaled by the different factors.

My question is whether or not piv is affected by these things.

I know well that even in NCPP case, psi is not meaningful near the nucleus of the atoms and I also guess that in PAW and USPP cases, we don't need to construct psi exactly considering augmentation parts. But, I am still wondering whether the enforcement of normalization in USPP and PAW cases works in all (or most) cases.

Sincerely,

Hyungjun Lee
EPFL SB Institute of Physics

giovannipizzi:
Hi, my guess is that the current implementation is (very often) ok (at least for the cases that we have tested). BTW, I think this is the same the authors of the original method did in their Matlab implementation. We're doing a more thorough validation study with more systems in the meantime.

@VVitale There's a typo in the code that prevents compilation I think: https://travis-ci.org/wannier-developers/wannier90/jobs/343964349#L600 could you fix it?

hjunlee:
It seems to me that there is a very minor bug in pw2wannier90:
In a serial case, the phase array has a wrong dimension:
phase(iw,iw)

Sincerely,

Hyungjun Lee
EPFL SB IoP

PR #168, opened 2018-02-14
edited the hamiltonian_write_rmn to correspond to hamiltonian_write_t…

stiegerc:
…b and use the proper formula for diagonal elements. For low nkpts or 2D structures, i.e. only 1 kpt in one direction, the old version produced false diagonal elements in the homecell, which should be equal to wannier centers in the wout file. Some testing revealed that the old version converges towards the new one with higher nkpts (or more vacuum) but for a 2D structure it'd be a complete waste adding more points just to fix the operator. I also added the number of R vectors on the 3rd line as it is in the hr hamiltonian file. Also removed the delta in hamiltonian_write_tb as the distinction of i==j or not is already taken care of by the if clause.


giovannipizzi:
Looks reasonable. I'm ok with the change of format (don't think many people are using it). For the change in formula, maybe @jryates and @ivosouza0 can have a more expert opinion on this.

giovannipizzi:
@jryates do you thing there is a setting file (and is it in the repo or in the codecov page?) to set some threshold for codecov so it does not show as failed for small changes? Also, I'm not sure how to interpret this 0%. (I know this does not prevent merging but still if it's below a threshold I think it's more intuitive also for us to see a green check)

PR #169, opened 2018-02-15
Removed deallocate spn from compute_orb, causing crash at compilation…

AntimoMarrazzo:
… time

(and actually not needed)


giovannipizzi:
Will be fixed with #167

PR #173, opened 2018-03-27
distribution of some large matrices in parallel run

hjunlee:
Dear developers:

Here are some modifications to distribute some large matrices in parallel run.

They come from my personal version of W90. Since it is more complex than I expected, I aligned the part of my version into the recent version of W90. But, with this version we can deal with the case which requires many k points and bands.

I assume that when gamma_only=T, we use only single core (it is natural since curretnly only k parallelisation is done in W90).

Sincerely,

Hyungjun Lee
EPFL


hjunlee:
I fixed minor bugs for the case with precond=T; Now all checks have passed.
(Since I am a novice to the github, I didn't know about the automatic checks.)

Sincerely,

Hyungjun Lee
EPFL

PR #176, opened 2018-05-20
Add possibility to print bvec.

sponce24:
- new input variable to trigger the printing
- new test


src/hamiltonian.F90
```diff
@@ -657,6 +657,23 @@ subroutine hamiltonian_write_rmn()
     end do
     
     close(file_unit)
+!  
+!   RM and SP - write to file vectors b and their weight wb for each k-point 
+!   This is used by EPW to compute the velocity. 
+!   Note that you need to put in your Wannier input:
+!     write_hr        = .true.
```
giovannipizzi: Instead of requiring all three flags to be set, could you put the code in a different function, and in `plot_main` (plot.F90) you do, toward the end, below
```
if(wannier_plot) call plot_wannier
```
something like
```
if(write_bvec) call plot_bvec
```

Note that the file will be changed when merging #177, this automatically also remove the need of `write_hr` (so you might want to work on top of those changes - in this case, add also the `.or. write_bvec` at the top, where the header is printed).

test-suite/tests/testw90_bvec/lead.win
```diff
@@ -0,0 +1,98 @@
+! Lead : Tutorial Example 2
+
+ num_wann        =   4 
+ num_iter        = 20 
+
+ write_hr        = .true.
```
giovannipizzi: I would make a test where only `write_bvec` is set, and all other plotting options are removed (`write_hr`, `write_rmn`) - see also my other comment above.

test-suite/tests/jobconfig
```diff
@@ -132,6 +132,13 @@ program = WANNIER90_WOUT_OK
 inputs_args = ('LaVO3.win', '')
 output = LaVO3.wout
 
+# Lead, 4 lowest states; Fermi surface, print bvec
+[testw90_bvec]
+program = WANNIER90_WOUT_OK
+inputs_args = ('lead.win', '')
+output = lead.wout
```
giovannipizzi: I think this test shouldn't check the .wout, but rather we should add a new `program` and a new `parser` to check the check the new `bvec` file (see e.g. what I did in #177 for a different file). You can then check with the parser, e.g., the number of bvecs or even all values.

giovannipizzi:
as a note, now #177 has been merged, I updated your branch so you can work directly on that version (see my comments above).

src/hamiltonian.F90
```diff
@@ -657,6 +657,23 @@ subroutine hamiltonian_write_rmn()
...
+!     write_rmn       = .true.
+!     write_bvec      = .true. 
+    if (write_bvec) then
+      open(file_unit,file=trim(seedname)//'.bvec',form='formatted',status='unknown',err=101)
+      write(file_unit,*) nntot
+      do nkp = 1, num_kpts
+         do nn = 1, nntot
+            write (file_unit,'(4F12.6)') bk(:,nn,nkp), wb(nn)
```
giovannipizzi: Another request about the format of the file: could you add two header lines, as in most of the other files:
1. first line as a comment, including a timestamp, in order to ease debug in case people regenerate many times the file and want to know if it was regenerated or not
2. Second line with the size of the matrix (two integers, `num_kpts` and `nntot`)

You can give a look to other files to see how the header looks like

PR #177, opened 2018-05-24
Fix U matrices not being written

### LINKED ISSUE #139: write_u_matrices not written
[`plot_main()`](https://github.com/wannier-developers/wannier90/blob/develop/src/plot.F90#L27) is called only in some cases. Some routines in it, therefore, are sometimes not called (e.g. `write_u_matrices`).

In fact, `write_u_matrices` is checked only inside `plot_main()`, but plot_main is called in [wannier_prog.F90](https://github.com/wannier-developers/wannier90/blob/develop/src/wannier_prog.F90#L229) only `if (wannier_plot .or. bands_plot .or. fermi_surface_plot .or. write_hr)`.
One needs to extend the if statements (or manage the plotting differently). (Plotting of course happens if at least one of the flags is also set in the input).

giovannipizzi:
This PR fixes #139 by removing the unneeded checks in wannier_prog for plotting.
These were used to decide hether to enter the plotting routines.
This was causing the `write_u_matrices` to be ignored if there wasn't also another plotting flag active.

I'm also adding a test to check that the `_u.mat` is actually being generated, and that at least the matrix shape is correct.


mostofi:
Looks good.

giovannipizzi:
Hi @mostofi, you closed instead of merging. I'll reopen it - you need to approve it and then merge it.

PR #189, opened 2018-06-28
Small fixes

jimustafa:
The first commit makes a small fix to README.install to reflect the change (#156) in the make target for running the tests.

The second commit removes trailing whitespace in the gitignore files throughout the repository. Trailing whitespace in the gitignore files was causing some files produced during testing not be ignored.

This brings up the question of how to deal with whitespace. It is common practice to exclude trailing whitespace, usually by configuring the text editor to trim whitespace upon saving. Perhaps it is desirable to trim trailing whitespace throughout the codebase and implement whitespace checking as part of accepting a pull request.

### DIFF
--- a/README.install
+++ b/README.install
@@ -48,7 +48,7 @@
                    the user guide)
    make w90vdw     build the van der Waals code
    make w90pov     build the ray-tracing code
-   make test       run test cases
+   make tests      run test cases
    make doc        build the documentation
    make dist       make a tar-ball of the distribution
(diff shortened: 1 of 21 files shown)


jryates:
I'd noticed that git-ignore wasn't working perfectly when cleaning the distribution - but it hadn't occurred to me that trailing whitespace was the issue! Thanks.
I'll open an issue so we can discuss your suggestion next time the developers get together.

Issue #131, opened 2017-04-03
Run with only parts of the projections given in the .amn file


greschd:
Dear all,

As far as I know, it's not currently possible to 're-use' the .amn file when changing the projections -- which means the first-principles code needs to be re-run. Since this can take a lot of time, I propose the following syntax to allow re-using the files:

* Add an optional input block (let's call it ``select_projections`` for now) with the same syntax as the ``projections`` block.
* If ``select_projections`` is not given, everything stays as it is
* If ``select_projections`` is given, the ``projections`` block is used to determine which projections are in the .amn file, and the ``select_projections`` block tells us which of these are actually to be used. 

This means that ``select_projections`` could be an arbitrary subset of ``projections`` without changing the .amn file.

Please let me know what you think about this.

mostofi:
Dear Dominik
To regenerate the projections, as long as you have saved the wavefunction
data, you should only need to calculate the overlaps and not the whole
electronic structure calculation again. In any case, this is an interesting
idea. There are some consistency checks that need to be thought about with
relation to the num_wann input parameter and depending on whether or not
the code is being run with the -pp command line option. Worth considering
though.
Best wishes,
Arash

greschd:
Dear Arash,
Thanks, I actually had not considered this. However, I'm still in favour of adding such an option, in particular since it can take quite a while for the first-principles code to start up and re-do the projections (especially for large systems). Also, I've found that there's quite a performance hit (at least in VASP) for keeping the wave functions.
If you don't mind, I'll try to create an implementation for this, together with @quanshengwu.
Best,
Dominik

greschd:
Hi Arash,
No, I haven't looked into this any further so far. I still think it could be useful, but it's not very high in priority for me.
Best,
Dominik

Issue #190, opened 2018-07-02
Trailing Whitespace


jryates:
From: @jimustafa and pull-request #189 
"This brings up the question of how to deal with whitespace. It is common practice to exclude trailing whitespace, usually by configuring the text editor to trim whitespace upon saving. Perhaps it is desirable to trim trailing whitespace throughout the codebase and implement whitespace checking as part of accepting a pull request."

giovannipizzi:
From a quick check, this looks like it's a good tool: https://pypi.org/project/fprettify/

If you agree @jryates @mostofi I will implement it in a test, and prepare a pre-commit hook, that can be installed locally, so that before committing the tool is run automatically for you (very easy to use).

Only thing to check would be to see if the FORD documentation is not "disrupted" but I don't think so, by default it only changes indentation/whitespaces.

giovannipizzi:
Long lines are apparently not supported, but we can replace
```
write(*,*) "aa&
  &bb"
```
with
```
write(*,*) "aa" // &
  "bb"
```
The plan is to wait that major PRs are merged, and then to try to apply this and see if there is no problem. Will be done in the next coding day.

giovannipizzi:
I also found this: https://github.com/dev-zero/f90-formatter
I think both are used by CP2K, I'll ask dev-zero (Tiziano Müller) what they suggest

pseewald:
@giovannipizzi 
The above link is an old fork of this repository:
https://github.com/pseewald/fprettify

You found this already on PyPi but currently only the github version is up to date (sorry, I will soon update PyPi release...).

CP2K uses fprettify and additionally a script for more aggressive formatting (removal of unused variables and sorting/reordering of variables and use statements). I don't recommend using the CP2K script since many of the formatting choices are CP2K-specific, also it is quite buggy, not maintained and will conflict with FORD.

If you give fprettify a try, please let me know if you encounter any problems.

PR #200, opened 2018-09-01
Select projections

jimustafa:
These changes are an attempt to address #131.

The primary change is the addition of some logic to remove the restriction that the number of projections must be equal to the number of Wannier functions. Now the projections block can define more than `num_wann` projections, as long as `num_proj` is defined. When the AMN file is being read, `select_projections` determines which elements enter in the construction of the projection matrices.

A few ancillary changes have also been made:
1. added a ["standard" TeX gitignore file](https://github.com/github/gitignore/blob/master/TeX.gitignore) for the documentation source tree
2. added magic comments to the user guide source TeX files to enable IDEs with TeX support to compile the document from any of the source files (eg, [Sublime Text](https://www.sublimetext.com/) using [LaTeXTools](https://latextools.readthedocs.io))

Additional remarks:
- specifying the `num_proj` parameter is a bit of a quiz for the user; it may be better to implement projections counting in `param_get_projections` (like in `param_get_range_vector`)


jryates:
Jamal - thanks for the interesting contribution. I can certainly see the logic in what you propose. I would like to talk through the implications with the other developers and will do so next time we have developers' call (we do this roughly every 6 weeks).
Jonathan

giovannipizzi:
Dear Jamal,
thanks a lot for your contribution, that looks very interesting.
We've discussed this at our developers' meeting and we would like to ask you to do a few changes before we merge. Could you do the following?

- [ ] rebase on develop (or merge and solve conflicts). The tricky part here is that we've introduced in #203 (see also #190) fixes for the indentation that changed most of the files. Luckily your changes to the `.F90` files are not extensive so in the worst case you can re-apply them on top of develop. After this, you'll find instructions on how to run the `fprettify` command to indent according to our code style here: https://github.com/wannier-developers/wannier90/wiki/ContributorsGuide#automatic-pre-commits-spaces-indentation- (sorry for the hassle...)
- [ ] print the output of the selected projections in the header: now only all of them are printed - good to print both all and the selected ones only https://github.com/wannier-developers/wannier90/blob/e233ffe3475e333d2347f00187c0bba9392b9f20/src/parameters.F90#L2582-L2600)
- [ ] currently, some parts of the code will not work/be compatible with your changes, because (e.g. in guiding centres and other places like selectively-localised WFs) the codes expects that `proj_site` (and the other projection-related arrays) have length `num_bands`. We suggest the following: 
  - Read the arrays in new arrays (e.g. `input_proj_site`, `input_proj_l`, ...)
  - after checking `select_projections`, fill in `proj_site` and all other projection-related arrays as expected by the rest of the code
  - in a few places, you will still need to use, instead, `input_proj_*`, like when printing the header (see point above) and in the functions in `kmesh.F90` when running in `-pp` mode to generate the `.nnkp` file.
- [ ] remove the `num_proj` from the input - as you note, it's redundant and potentially confusing. This can be set internally from the `select_projections` input.

Thanks a lot in advance!

jimustafa:
Thanks for the feedback! I see your points and have tried to make the appropriate changes.

The last point is something that I still have been unable to address. Since the input_proj_* arrays are allocated within param_get_projections, it is not clear to me how to know the number of projections beforehand. It seems that counting the number of projections within param_get_projections, and then calling the routine again after num_proj has been set, is not that straightforward.

Please advise me on how to best determine num_proj before a call to param_get_projections. Much appreciated!

giovannipizzi:
There a few ways that come to my mind to do this.
The approach that to me looks closer to what is done for other parts in the code (e.g. when reading blocks) is the following:
- add a boolean flag to `param_get_projections` (`count`, that if True will just return the value), and an additional integer that will be set when `count = .true.` (or actually could always be set to num_proj - but you still need the count=True boolean flag to avoid allocation in the first round).
- Add a few 'if' statements in the `param_get_projections` routine to skip parts that would allocate the `proj_*` arrays at the beginning (so that this is not done when `count = True`), and of setting the arrays. In the meantime keep a counter of how many projections there are.
- As soon as possible, if `count = True`, set the additional integer parameter to the detected value on `num_proj` and return (e.g. skipping the final normalisation).

Actually, the additional flag could have an intent inout and be used in the second call, with count=False, to pass in the value of `num_proj`.

giovannipizzi:
Hi @jimustafa, just to update on our internal timeline:
- we met today for a Wannier90 coding day
- we will meet again end of November, so if you have the chance to do the remaining changes before Nov 25 it would be great, so we have the time to review and merge it. If you don't think you'll manage, please let us know so next month we'll merge your PR in a branch and continue working on it
- we plan to release 3.0 around mid-December

Thanks!

jimustafa:
With the most recent commit, I believe the feature has been completed. If there are any additional items identified during the next Wannier90 coding day, please let me know and I will try to have them complete by mid-December. Any feedback from @greschd would be appreciated, regarding whether the feature helps with the use cases that led to #131.

greschd:
Hi @jimustafa. I haven't had the chance to try this out yet, but from what I see it does indeed seem to solve the use case for #131. Thanks a lot!

PR #205, opened 2018-10-08
Minor changes to SLWF+C logic in parameters.F90

VVitale:
to better handle tricky situations when combining SCDM with SLWF+C

### DIFF
--- a/src/parameters.F90
+++ b/src/parameters.F90
@@ -809,10 +809,15 @@ contains
     slwf_constrain = .false.
     call param_get_keyword('slwf_constrain', found, l_value=slwf_constrain)
     if (found .and. slwf_constrain) then
-      allocate (ccentres_frac(num_wann, 3), stat=ierr)
-      if (ierr /= 0) call io_error('Error allocating ccentres_frac in param_get_centre_constraints')
-      allocate (ccentres_cart(num_wann, 3), stat=ierr)
-      if (ierr /= 0) call io_error('Error allocating ccentres_cart in param_get_centre_constraints')
+      if (selective_loc) then
+        allocate (ccentres_frac(num_wann, 3), stat=ierr)
+        if (ierr /= 0) call io_error('Error allocating ccentres_frac in param_get_centre_constraints')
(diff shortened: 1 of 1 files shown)


src/parameters.F90
```diff
@@ -2445,7 +2449,7 @@ function get_smearing_index(string, keyword)
 
   end function get_smearing_index
 
-  !===================================================================
+!===================================================================
```
giovannipizzi: Interesting that this indentation changed... I wasn't expecting this to happen.

src/parameters.F90
```diff
@@ -2159,28 +2165,26 @@ subroutine param_read()
...
+      ! Check that either projections or constrained centres are specified if slwf_constrain=.true.
+    elseif (slwf_constrain .and. .not. (found .or. allocated(proj_site))) then
+      call io_error('Error: slwf_constrain = true, but neither &
+           & <slwf_centre> block  nor &
+           & <projection_block> are specified.')
     else
-      if (selective_loc .and. slwf_constrain .and. allocated(proj_site)) &
-        write (stdout, '(a)') 'No slwf_centres block requested. Desired centres for SLWF same as projection centres'
+      ! Allocate array for constrained centres
+      call param_get_centre_constraints
```
giovannipizzi: Shouldn't there be a check that `selective_loc` is true? The function internally accesses variables that are now allocated only `if (selected_loc)` above

VVitale: Hi @giovannipizzi, I changed it so it is easier to follow. I don't think I need a check on `selective_loc`, since `slwf_constrain` is true only if `selective_loc` is also true.

VVitale:
Yes, I think that was the case in my previous PR. It should be OK now, all the tests have passed.

mostofi:
Seems fine to me. @giovannipizzi - I think your concern has been addressed, and this PR can now be accepted?

giovannipizzi:
Yes, I had checked it when the tests were still running. Now I merged it!

PR #215, opened 2018-11-27
Fixing a small bug introduced in #200

giovannipizzi:
When lcount=.true., some content of the
input file that was read was changed, and
the second run (e.g. if 'random' was present)
was giving wrong results.
Not clear because the error was errand, though
(was working on travis but failing on my machine)


giovannipizzi:
The input file is stored in a list of strings (and slightly modified to remove comments, fix capitalisation etc.)
While before the `param_get_projections` was called only once, so it wasn't a problem if it was modifying the list of strings, now it's called twice (once to count and once to get them) so it shouldn't change the input array.

Issue #210, opened 2018-11-01
Issue with ws_distance


giovannipizzi:
We (@AntimoMarrazzo and me) have a problem with a specific input.

The code prints `ERROR in hamiltonian_wigner_seitz: error in finding Wigner-Seitz points`.

There are two issues here:
1. following the code path, I think the routine is always called, also when `use_ws_distance` is False. Maybe we should skip that part, so that if an error occurs, at least the code does not stop when setting `use_ws_distance=.false.`.
2. I don't know why the code stops. I attach here the input [tmp.win](https://github.com/wannier-developers/wannier90/files/2537934/tmp.win.txt) and the output with high-verbosity 
[tmp.wout](https://github.com/wannier-developers/wannier90/files/2537935/tmp.wout.txt) (rename the files removing `.txt` if you want to run again). There are many kpoints so it is hard to check "by eye".
(it finds `tot = 3524.00000000000`, while the number of points in the `mp_grid` is `3528`).
We even tried to increase from a 2x2x2 supercell to a larger one and to change a bit one of the thresholds, but apparently this didn't work in the sense that tot is always equal to 3524 (but we should re-check to be sure).

Note: to reproduce without running the DFT calculation, you can just do the following changes:
- in `hamiltonian.F90`, add `public :: hamiltonian_wigner_seitz` at the top and declare `ham_k` as public: `complex(kind=dp), save, allocatable, public :: ham_k(:, :, :)`
- in `wannier_prog.F90`:
  - add `integer :: ierr` at the top
  - add the following block inside `if (postproc_setup)`:
    ```fortran
    print *, '### DEBUG ###'
    if (on_root) call hamiltonian_wigner_seitz(count_pts=.true.)
    allocate (irvec(3, nrpts), stat=ierr)
    if (ierr /= 0) call io_error('Error in allocating irvec in hamiltonian_setup')
    irvec = 0
    !
    allocate (ndegen(nrpts), stat=ierr)
    if (ierr /= 0) call io_error('Error in allocating ndegen in hamiltonian_setup')
    ndegen = 0
    !
    allocate (ham_r(num_wann, num_wann, nrpts), stat=ierr)
    if (ierr /= 0) call io_error('Error in allocating ham_r in hamiltonian_setup')
    ham_r = cmplx_0
    !
    allocate (ham_k(num_wann, num_wann, num_kpts), stat=ierr)
    if (ierr /= 0) call io_error('Error in allocating ham_k in hamiltonian_setup')
[...]

Issue #39, opened 2016-09-15
Make the translate_home_cell work also when plotting the WF


giovannipizzi:
Now, it is not applied while plotting and while printing the actual WF centres on the output

jryates:
At the moment we translate only the centres, and this leaves the output (e.g. U matrices, r^2, H_r) inconsistent. This is an issue for PythonTB see 
http://www.physics.rutgers.edu/pythtb/usage.html#pythtb.w90

VVitale:
At the moment there are two places in the code where `translate_home_cell` is used: 1) when writing the cartesian coordinates of the WF's centres and 2) when writing VdW data (which also prints out the WF's centres). 
The translations are done in a naive way by directly modifying the centres of the WFs without modifying the U matrix. As Jonathan and Giovanni pointed out, this makes the output of other quantities, e.g. r2_mn, U matrix, H matrix and the plotting of WFs inconsistent, which is a problem for TB codes.

By directly modifying the columns of the U matrix with the appropriate phase factors one can make all the other quantities consistent with the translation of the centres (I haven't thought about the analyticity of the resulting U matrix, but I don't think it should be a problem though). In this case we can imagine several solutions

1) We modify the original U matrix directly at the end of the minimisation
(what happens when restarting? discontinuity?)
2) We only create and modify a copy of the U matrix when printing in output if 
`translate_home_cell = .true.` (Personally I would go for this option, as it looks neat)
 3) We do not modify the U matrix but only modify the quantities (r2mn, H etc.) when printing in output.
This option has the advantage that we do not need to modify the U matrix, but it has the disadvantage
that we have to modify all the plotting and printing routines instead.

There is also another routine `internal_translate_centres` in the `hamiltonian.f90` module that performs basically the same operation as `utility_translate_home`. This routine is only called in the following cases

    if ( bands_plot .and. (index(bands_plot_mode,'cut').ne.0) ) use_translation=.true.
    if ( transport  .and. (index(transport_mode,'bulk').ne.0) ) use_translation=.true.
    if ( transport  .and. (index(transport_mode,'lcr' ).ne.0) ) use_translation=.true.

We are not sure if we should maintain this "copy" or only use one of the two routines in the code.

VVitale:
We need some more testing to make sure that the centres and the Hamiltonian matrix elements computed by applying a phase to the columns of the U matrix agree with the way we computed them now (i.e. manually translating the centres of the WFs).

PR #216, opened 2018-12-02
Fixed issue in the search for points inside the Wigner-Seitz cell

gibertini:
This commit should fix this issue:
https://github.com/wannier-developers/wannier90/issues/210#issue-376332272

When unit cells are very anisotropic (e.g. very stretched in one direction) the original range of points among which the code searches for points inside the Wigner-Seitz cell was insufficient. 

The range is now larger to comply with these cases and it's no longer hard coded, so that in case of need it can be further enlarged by changing the variable search_size in hamiltonian_wigner_seitz.


src/hamiltonian.F90
```diff
@@ -578,18 +583,21 @@ subroutine hamiltonian_wigner_seitz(count_pts)
...
+      tot = tot + 1.0_dp/real(ndegen(i), dp)
+    enddo
+
     if (iprint >= 3 .and. on_root) then
       write (stdout, '(1x,i4,a,/)') nrpts, ' lattice points in Wigner-Seitz supercell:'
       do i = 1, nrpts
         write (stdout, '(4x,a,3(i3,1x),a,i2)') '  vector ', irvec(1, i), irvec(2, i), &
           irvec(3, i), '  degeneracy: ', ndegen(i)
       enddo
+      write (stdout, *) ' tot = ', tot
```
giovannipizzi: Add an explicit format as in the lines above

src/hamiltonian.F90
```diff
@@ -505,18 +505,19 @@ subroutine hamiltonian_wigner_seitz(count_pts)
 
     logical, intent(in) :: count_pts
     !! Only count points and return
-
+    integer, parameter :: search_size = 2
```
giovannipizzi: As we discussed, make search_size an input parameter, and make a quick benchmark to see, with a dense k-mesh, if 2 is much slower (then leave 1 as default) or 2 is ok.

gibertini:
Following the suggestions by @giovannipizzi I introduced a new input variable (`ws_search_size`). 
This variable is now used consistently in all routines that involve the search for points inside the Wigner-Seitz cell (that is inside `hamiltonian.F90` and `ws_distance.F90`). I have done something similar also for `ws_distance_tol` that is now used also in `hamiltonian.F90`. 
The default value for `ws_search_size` has been set to 2. This does not introduce a serious slow-down of the code, and is safer when using `use_ws_distance = .true.` as the WF centres can wander around in neighbouring unit cells (the previous default in that part of the code was 3). Maybe it would be good to have a feedback from @paulatz on this choice.

giovannipizzi:
Indeed, two questions we need to answer (after discussion with @gibertini today) are:
- is it ok to use the same variable for both? (and is the value a good compromise)?
- Same question after #39 is merged? (even if people might not be using translate_home_cell, and probably we will implement it only for plotting and not for the actual matrices? mentioning also @VVitale for feedback)

If we are unsure, possibly the best approach is to put two different variables, setting the defaults to 2 and 3, respectively. We just need to find good names.

gibertini:
As a compromise, the local value of ws_search_size is incremented by one inside ws_distance.F90.
I added a comment regarding this, both in the source code and in the documentation.

Issue #222, opened 2018-12-20
changed behaviour of guiding_centres  in recent versions


gibertini:
When considering cells with a large amount of vacuum and atomic positions around the centre of the cell, the minimisation procedure sometimes has issues and the spreads become quite large.
This applies for instance to 2D materials that extend along the xy plane and are centred at z = 0.5 (in crystal units).
I was always able to solve this problem in the past by using "guiding_centres = true".
I noticed that this is no longer the case in recent development versions.
In particular the latest versions (irrespective of the value of guiding_centres, true or false) provide the same wrong result as the old versions with guiding_centres = false, as if its effect would be now "disabled". 

I'm providing here a very simple example that shows this issue ([test_guiding_centres.tar.gz](https://github.com/wannier-developers/wannier90/files/2699903/test_guiding_centres.tar.gz)).
It's a one-dimensional chain of 10 Na atoms, sampled at gamma only. 
Up to version 2.1 the code finds a good minimum, with reasonable spreads. In the development version instead the spread is unphysically high.

I tried to perform a check on when the change in the behaviour was introduced using git bisect and this script ([test_revision.txt](https://github.com/wannier-developers/wannier90/files/2699859/test_revision.txt)).
I simply ran
`git bisect start`
`git bisect bad develop`
`git bisect good v2.1`
`git bisect run ./test_revision.sh`
but unfortunately many commits are skipped because the code ends with a segmentation fault when running my example, so that I'm left with more than 40 commits that cannot be further bisected:
a6706bab17b4fd42c9ce403c30dd366eb621cff8
66e8d5b624179eb55338fd15e70b3a923373673d
fb629b112892b4772fee3c2ca37244ae0e0b0f30
01786409ada479d3140892d11e3366f4caa3fc6c
6b424608b23c3f5e9de880c5e0fcf8d70b5a2ac3
b04258a2aa408965f5883384dd140de5487f24f2
fb96501c5a8680d961efff89b543a0427bafd826
00799a4fcb73f1ae6e3303d16d7a7561e8c0349e
aaf6b22ec83c7c5dd7fdc17ca73648202ef75874
2cd876e2d60b42e96fe24aa75b5261a7b0a183f3
af87eb87107183a875947dc1eec79f3052fba3c7
48050b8b4f9e6ff378f9e9404d7e01fea2e30dba
b98bd7be70091812e5407758cc8db47212a6dea7
0f2a17a6f2370215bc35d289b175e73f84facd32
[...]

PR #223, opened 2018-12-21
Fix guiding center gamma point bug

jryates:
Fixes issues #222 Also adds regression test for this bug


src/disentangle.F90
```diff
@@ -533,7 +533,8 @@ subroutine internal_find_u_gamma()
 
       u_opt_r(:, :) = real(u_matrix_opt(1:ndimwin(1), 1:num_wann, 1), dp)
 
-      a_matrix_r(:, :) = real(a_matrix(:, :, 1), kind=dp)
+      write (77, *) size(a_matrix_r, 1), size(a_matrix, 1)
```
giovannipizzi: Probably this write is a leftover?

test-suite/tests/testw90_na_chain_gamma/fort.77
```diff
@@ -0,0 +1 @@
+          21          30
```
giovannipizzi: Probably this file is not needed and can be removed?

PR #238, opened 2019-02-08
Move SCDM flags to pw2wannier90

### LINKED ISSUE #237: Move SCDM input keywords from Wannier90 to pw2wannier90?
@jryates @mostofi @giovannipizzi 

Pros:
* Makes the implementation more consistent. It would avoid writing the scdm_info block in the .nnkp file (restoring back compatibility).
* More easy to generalise for other codes. They won't have to modify Wannier90 if a better or just different implementation of SCDM pops up in the future.
* From the user perspective doesn't change much, the four scdm keywords will seamlessly be moved to pw2wannier90.

Cons:
* Need to change the codes now before release and documentation too.
* Need to change the AiiDA workflow
* Need to be tested again to avoid introducing new bugs.

VVitale:
This fixes #237. 
Moving the four SCDM flags from parameters.F90 to pw2wannier.F90 (now in pwscf/POST-v6.3-with-SCDM folder). 
No scdm_info block is written into .nnkp files for back-compatibility. 
**New behaviour**:
1) If there is a projection block in the .win file then the projection block in the .nnkp file remains the same . In this case if scdm_proj = T in the .pw2wan input file the code uses the SCDM method with the given number of projections. A warning is fired both in the pw2wannier90 output file and wannier90 output file to make the user aware of this. The warning also provides some suggestions on how to change the input files if this is not the intended behaviour.
2)  If no projection block is present in the wannier90 input file, then the projection block in the .nnkp file looks like this
```
begin projections
     0     num_wann
end projections
```
Where `num_wann` is the number of projections to use  in the SCDM. I have code this in such a way that doesn't break back compatibility.

I have also modified kmesh.F90 such as no scdm_info block is written into .nnkp files; overlap.F90 in order to read the new header in the .amn files and set the number of projections accordingly. 

I have also made changes to the input files of example27 and modified the tutorial. I have removed the description of the scdm keywords from the user guide but I have left the section describing the SCDM method.


VVitale:
@giovannipizzi has made the following suggestion to make the code fully back-compatible and also in a certain sense forward-compatible with the Mustafa implementation (selection of fewer projections):
We write a block in the .nnkp code with the following structure

```
begin partial_projections
  tot_num_wannier
  0 # or number of following lines 
  # every line is an index of the projection in the projections block to use for partially-automated
end partial_projections
```
or something similar. In this way we pass the info about the number of projections to the SCDM in back compatible way.  This block is only written if a specific **new** flag is `T` in the wannier90 input (e.g. `use_auto_proj`), so that wannier90 "knows" that the SCDM has been used through an internal flag and not only through the header of the .amn file (which is not very safe).  We then have four cases that we can handle safely and the logic can be extended to future implementations when fewer projections than the number specified in the projection block are selected.
@mostofi @jryates what do you think? Do you see any flaw in this logic? Do you have any preference for the name of the new flag and the block in .nnkp?

giovannipizzi:
After discussion, we'll call the flag in Wannier90 `auto_projections` (boolean) and put a block `begin auto_projections` in the `nnkp` file, only if the flag is true. @VVitale will implement correct warning/failure logic depending on the combination of flags both in Wannier90 and pw2annnier90, to minimise the risk of a user thinking to do something but actually computing something else.

PR #245, opened 2019-02-25
Making sure the distribution tar.gz is slim

### LINKED ISSUE #243: Improve make dist
- The previous `make dist` requires to be updated
- The current one (using `git archive`) creates a too big file

Idea: use `git archive` but excluding some folders (notably the test folder)? (check if it's possible), i.e. blacklisting file rather than whitelisting them

giovannipizzi:
In particular, avoiding to include the test-suite folder
that is quite big. 

Fixes #243


giovannipizzi:
closing in favour of #250

PR #256, opened 2019-03-05
Solution_booklet Errata from Alex Durie

VVitale:


VVitale:
Yes me too. Not sure why I used the hyphen in that paragraph.

PR #258, opened 2019-03-25
Fix a bug and add a check in parameters.f90

qiaojunfeng:
1. It seems that the `berry_task` check for `morb` is wrong in `parameters.f90`, while in the previous version of `wannier90` this was correct. 

2. I have added a check for `kpoint_path` block since I accidentally encountered a segmentation fault when plotting kpath. The situation is that if `kpath = true` while there is no `begin kpoint_path ... end kpoint_path` block in the input file, the `postw90.x` will complain: 

        Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

        Backtrace for this error:
        #0  0x7f63e4604dff in ???
        #1  0x55cc0cdecdd0 in ???
        #2  0x55cc0cd444af in ???
        #3  0x55cc0cd44871 in ???
        #4  0x7f63e45f1222 in ???
        #5  0x55cc0cd4349d in ???
        #6  0xffffffffffffffff in ???
        [1]    14953 segmentation fault (core dumped)  ../../postw90.x Fe

    compiled with gfortran 8.2.1 on Linux.

3. I recently implented a new feature of calculating spin Hall conductivity based on wannier90 (https://journals.aps.org/prb/abstract/10.1103/PhysRevB.98.214402). I would like to share my code with the community. If it could be merged into wannier90, I would like to write the corresponding documentaion and tutorials and create a pull request in the next few weeks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@ Moreover, it generates a new file `seedname_band.labelinfo.dat` with information
 
 - Improvements to the Wigner-Seitz detection routines [[#117]](https://github.com/wannier-developers/wannier90/pull/117) [[#109]](https://github.com/wannier-developers/wannier90/pull/109)
 
+- Fix berry_task check for morb, and add check for kpoint_path block in parameters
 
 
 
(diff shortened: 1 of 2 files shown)


jryates:
Thanks for this. Yes - if you could make a pull request for the spin Hall code that would be great. Post an issue if you would like to discuss any of the details.

PR #264, opened 2019-04-10
New feature for postw90: spin Hall conductivity

qiaojunfeng:
Hello, I have finished the code and documentation for calculating spin Hall conductivity. 

First I would like to thank you Wannier developers for open-sourcing such a great tool, for deriving the theories of maximally localized Wannier functions and expanding its usages to the calculations of lots of physical properties. 

Then some words about my work. The idea is similar to that of calculating anomalous Hall conductivity/optical conductivity: using Wannier interpolation to calculate some matrices needed by Kubo formula. The detailed description can be found in my paper [Phys. Rev. B 98, 214402 (2018) ](https://doi.org/10.1103/PhysRevB.98.214402). Two examples (example 29 & 30) are added, two test-suites are added, and the related user guide/tutorial/solution booklet are updated as well.

By the way, I have also fixed some small typos & errors in the documentation and test-suite, sorry that I didn't open a sperate pull request cause they are very tiny bugs.

Hope this can be useful for those interested in this topic. Thank you.


jryates:
Thanks! It might take me a week or so to take a proper look at your code before I can merge it.

I think your test is failing because you have a relative error set for a parameter which can be close to zero - you can set an absolute error instead - see https://testcode.readthedocs.io/en/latest/userconfig.html#tolerance

PR #265, opened 2019-04-13
[bug fix] Extension maximum length of file_name in kpath and kslice in postw90

qhapaq-49:
I solved a small problem with postw90. Maximum length of output file name (eg, python script for plotting) is too small. In the current kpath.F90, only 20 characters can be used as a file name including a suffix (eg "-bands + curv_z.py"). It causes trouble when one tries to make seed name longer than 5 like BaTiO3.

I thank all of wannier90 developers.


jryates:
Thanks for spotting and fixing this!

Issue #269, opened 2019-04-29
Problem (precision) with the SHC tests


giovannipizzi:
The new SHC tests (introduced in #264 by @qiaojunfeng) are failing on some compilers, see:

http://130.186.13.169:8010/#/builders/15/builds/20/steps/5/logs/stdio

http://130.186.13.169:8010/#/builders/12/builds/26/steps/5/logs/stdio

http://130.186.13.169:8010/#/builders/18/builds/22/steps/6/logs/stdio

Is this expected (and increasing the threshold in the tests is ok) or does it hint at a bug?

qiaojunfeng:
It seems that all the errors are less than 1e-2 relative to the benchmark values. I don't have much experience in these precision issues of compilers, but from a physical point of view, these results are acceptable.

giovannipizzi:
Fixed in #273

giovannipizzi:
I'm actually reopening this - there are still two tests failing:
- http://130.186.13.169:8010/#/builders/18/builds/26
- http://130.186.13.169:8010/#/builders/7/builds/27

PR #273, opened 2019-05-28
Two bugs in pw90common_fourier_R_to_k*

### LINKED ISSUE #271: use_ws_distance in pw90common_fourier_R_to_k_vec
When using the `pw90common_fourier_R_to_k_new` and `pw90common_fourier_R_to_k_vec` in `postw90_common.F90`, I noticed they produce different results, which is caused by the repeated lines in `pw90common_fourier_R_to_k_vec`.
https://github.com/wannier-developers/wannier90/blob/280f38267410f085244732a1577f09433fa1500f/src/postw90/postw90_common.F90#L1083-L1086

In the `use_ws_distance = .true.` branch, the `rdotk` and `phase_fac` are calculated twice. The first two lines use the new Wigner-Seitz cell interpolation, while the last two lines use the old interpolation. As a result, the old interpolation replaces the Wigner-Seitz cell interpolation. The last two lines should be deleted.

There are repeated lines in `pw90common_fourier_R_to_k_vec_dadb` as well.
https://github.com/wannier-developers/wannier90/blob/280f38267410f085244732a1577f09433fa1500f/src/postw90/postw90_common.F90#L1170-L1173
 
 
The SHC code uses both `pw90common_fourier_R_to_k_new` and `pw90common_fourier_R_to_k_vec`
https://github.com/wannier-developers/wannier90/blob/280f38267410f085244732a1577f09433fa1500f/src/postw90/berry.F90#L1959
https://github.com/wannier-developers/wannier90/blob/280f38267410f085244732a1577f09433fa1500f/src/postw90/berry.F90#L1966
When I change the wrong `pw90common_fourier_R_to_k_vec` to the correct `pw90common_fourier_R_to_k_new`, the `testpostw90_pt_shc` test case produces large deviations, 
```
shc
    ERROR: absolute error 7.05e+01 greater than 1.00e-03. (Test: 280.60944.  Benchmark: 351.10919.)
    ERROR: relative error 2.01e-01 greater than 2.00e-03. (Test: 280.60944.  Benchmark: 351.10919.)
shc
    ERROR: absolute error 7.12e+01 greater than 1.00e-03. (Test: 282.67215.  Benchmark: 353.83371.)
    ERROR: relative error 2.01e-01 greater than 2.00e-03. (Test: 282.67215.  Benchmark: 353.83371.)
```
Fortunately, when using a denser `berry_kmesh` like `100 * 100 * 100`, the final result doesn't change much compared to previous calculations:
pt_diff
The `Pt-shc-fermiscan.dat.old.10` and `Pt-shc-fermiscan.dat.old.100` are calculated with the `develop` branch, the `Pt-shc-fermiscan.dat.new.10` and `Pt-shc-fermiscan.dat.new.100` are calculated with [my modified branch](https://github.com/qiaojunfeng/wannier90/tree/shc_mem), which uses `pw90common_fourier_R_to_k_new` instead of `pw90common_fourier_R_to_k_vec`.

When this problem is fixed, I may need to update benchmark files in the SHC test cases.

qiaojunfeng:
There are 2 major bugs in routines `pw90common_fourier_R_to_k*`

1. The loop index `i, j` on `wdist_ndeg` is wrong, and this will cause wrong interpolation results of all the calculations. In some bandstructure calculations, the Wannier interpolated bands significantly deviate from ab initio bands. Applying fixes of this commit can correctly interpolate bandstructure. 

2. The repeated calculations of `rdotk` and `phase_fac` can lead to wrong results, as pointed out in issue [271](https://github.com/wannier-developers/wannier90/issues/271).

Fixes #271.


giovannipizzi:
Thanks! Indeed the loops seem wrong. Do you have an example where the interpolation is wrong because of this bug and gets much better after the fix?

qiaojunfeng:
The attachment is a zip file including the input and output files of QE & wannier90.
[input_output.zip](https://github.com/wannier-developers/wannier90/files/3358249/input_output.zip)

The structure of the zip is:
```
./input_output
├── bands.in
├── bands.x.in
├── Bi.rel-pbe-dn-kjpaw_psl.0.2.2.UPF
├── figure
│   ├── new
│   │   ├── mbt-bands.dat
│   │   ├── mbt-bands.gnu
│   │   ├── mbt-bands.pdf
│   │   ├── mbt-bands.py
│   │   └── mbt-path.kpt
│   ├── old
│   │   ├── mbt-bands.dat
│   │   ├── mbt-bands.gnu
│   │   ├── mbt-bands.pdf
│   │   └── mbt-bands.py
│   └── qeband.pdf
├── mbt.pw2wan
├── mbt.win
├── mbt.wout
├── Mn.rel-pbe-spn-kjpaw_psl.0.3.1.UPF
├── nscf.in
├── scf.in
└── Te.rel-pbe-dn-kjpaw_psl.0.3.1.UPF
```
To reduce computational time, I have decreased energy cutoff, kmesh in QE run, and the number of iterations in wannierisation step, so the spreads of MLWF are not very good, around `3.8`, as shown in `./input_output/mbt.wout`. 

The file `./input_output/figure/qeband.pdf` is the ab initio band produced by `plotband.x` of QE.
The file `./input_output/figure/old/mbt-bands.pdf` is the Wannier interpolation band produced by the `develop` branch of `postw90.x`.
The file `./input_output/figure/new/mbt-bands.pdf` is the Wannier interpolation band produced by my modified `postw90.x`. 

Since the `chk, amn, mmn, eig` files are too large, I haven't included them in the tarball. 

For your convenience, the figures are pasted here:
qeband
old
new

stepan-tsirkin:
Dear Junfeng, thank you for pointing to this bug. 

I also suspect there is another bug with ws_distance in postw90_common.F90

line 732 ( subroutine pw90common_fourier_R_to_k )
reads:
              OO(i, j) = OO(i, j) + cmplx_i*crdist_ws(alpha, ideg, i, j, ir)*phase_fac*OO_R(i, j, ir)

which I believe is correct. However in the other versions of this subroutine the array crvec is used instead of crdist_ws, also when use_ws_distance=True

see lines 732, 820-825, 881, 888, 995, 1093-1101, 1182, 1318

e.g line 820:
        if (present(OO_dx)) OO_dx(:, :) = OO_dx(:, :) + &
                                          cmplx_i*crvec(1, ir)*phase_fac*OO_R(:, :, ir)

Am I right, that those lines also need to be corrected (crvec-> crdist_ws) ?

This does not change the interpolated energies, but plays a role when interpolating velocities, Berry curvatures, etc

gibertini:
Dear Junfeng, 

thanks for spotting this. I also totally agree that both 1. and 2. are bugs and I believe that your commits are the right solution.

For 1. indeed the indices `i` and `j` of `wdist_ndeg` are not used consistently within the same loops. So for sure one needs to make a consistent choice. Among the two possibilities (`i,j` or `j,i`), I agree that the one you chose is the correct one.  

For 2. I think the re-definition of `rdotk` and `phase_fac` is the result of an unfortunate cut-and-paste from the old version that was not commented out. Indeed, whenever `rdotk` and `phase_fac` are defined in the new way, the old version is kept right below and commented. I think it would be simply better to remove all such commented lines, together with the ones erroneously uncommented that you spotted.

gibertini:
Dear Stepan,

It think you are right. There was initially an issue with the interpolation of band velocity that was solved in #53 .
Unfortunately the solution was not extended to other interpolation routines.

Thanks you for spotting this!

stepan-tsirkin:
Good, How do we proceed? 
 Should I open a separate pull request to implement what I noted in the previous comment?
Or just the current PR may be updated.?

qiaojunfeng:
Dear Marco Gibertini, 

Thanks! I have removed those comments.

qiaojunfeng:
Dear Stepan S. Tsirkin, 

Thanks! I think it should be `crdist_ws` too. If you like, I can help fix the lines you mentioned, or you fix the bugs by creating a new pull request, which do you prefer?

stepan-tsirkin:
Dear  Junfeng,
It would be nice if you fix those lines, and I double-check afterwards. 
It is a related bug, so it does not deserve a separate PR, I think.

qiaojunfeng:
Dear Stepan,

I have corrected the `crdist_ws` bugs, could you please double-check my modifications?
I split some long lines to ensure their length < 132, otherwise, the compiler will complain:
```
../postw90/postw90_common.F90:1001:132:

 1001 |                                         (crdist_ws(a, ideg, i, j, ir) + local_wannier_centres(a, j) - local_wannier_centres(a, i))* &
      |                                                                                                                                    1
Error: Line truncated at (1) [-Werror=line-truncation]
```

stepan-tsirkin:
Hi Jungeng,

Thank you, the correction look fine.

Just now after a careful look, in lines 1302 
and 1312 it seems to me that irvec also should be replaced to irdist. But to be sure it is better to consult the contributors that implemented that ..._TB_conv subroutine

giovannipizzi:
If I am not mistaken, those lines are introduced [here](https://github.com/wannier-developers/wannier90/commit/f602c55989372cc28bea8d57377ae4fd49d7cab8#diff-3cf0ff4c5b67090202268b7b6a353c6eR1295) by @Julen-ia but they have comments with `[lp]` who I guess is @paulatz (maybe copy-paste from elsewhere?).

Could @Julen-ia and @paulatz comment on these points/possible bugs? (We are having a coding day on the 24th and if we figure out everything before then we can review and merge). Thanks!

Julen-ia:
Dear all,

Thank you for spotting the bugs, and also these two lines that I introduced and need to be changed. I think that the safest solution is to specify that both irdist_ws and irvec need to be zero in the if statement Stepan spotted.

In principle, the if statement should only apply to one R vector, namely \vec{R}=0 (the origin). In ws_distance.f90, line 136, I see

irdist_ws(:, ideg, iw, jw, ir) = irvec(:, ir) + shifts(:, ideg)

if we switch irvec --> irdist_ws in the if statement Stepan spotted, irdist_ws = 0 could be triggered even if irvec(:,ir) \neq 0 provided irvec = - shifts , and this is not what we want. I dont have much experience with irdist_ws, so what I just said may never take place. Unfortunately I am on travel at the moment and I have limited availability for doing numerical checks.

Any feedback is appreciated, thank you all

Best regards

Julen

paulatz:
It's been a while, and I'm not sure I remember everything correctly. From what I see, the "old" subroutine  pw90common_fourier_R_to_k should be correct, except for the exchange of i with j in wdist_ndeg. I think this is not really a bug since wdist_ndeg(i,j,ir) is symmetric in i,j (I cannot think of any case where it could be non-symmetric, but I may be wrong).

On the other hand, the "new" subroutine pw90common_fourier_R_to_k_new was actually still using the "old" interpolation, without using the minimum distance. All the changes look correct to me and should fix the problem.

In pw90common_fourier_R_to_k_vec (line ~1087) the re-definition of the phase factor also looks quite clearly like a bug to me and I think the new code fixes it.

cheers

qiaojunfeng:
Dear Julen, 

Thanks for the response! 
I spent some time reading [your PRB paper](https://doi.org/10.1103/PhysRevB.97.245143), and I will try to convey my thoughts about this `if` statement. 

From Eq.(21) in your paper, the TB convention is 


In the Wigner Seitz interpolation, we replace the ordinary interpolation

by

i.e. the original `R` is replaced by `R+T` to ensure minimal distance between Wannier centers `r_m` and `r_n + R + T`, where `T` is the supercell translations. (These equations can be found in the v3 review paper Eq.(47,48) ).

Since `R+T` is identical to `R`, the TB convention should also hold for any `0 + T`, not only `R = 0`. 
For `ir` in the range of `1` to `nrpts`, there exists exactly 1 `ir` which satisfy `irvec(:,ir) = 0`, that is this `ir` corresponds to `R = 0`. For this `ir`, the TB convention is satisfied, and its translations `R + T` should also satisfy TB convention, i.e. all the `irdist_ws(:, ideg, i, i, ir)` with `ideg` ranges from `1` to `wdist_ndeg(i, i, ir)`. So, if we specify both `irdist_ws` and `irvec` be `0`, then we may lose some `irdist_ws` which are the degenerate of `R = 0`. 
In reverse, is it possible that for some `ir`, `irvec(:,ir) != 0` but `irdist_ws(:,ideg,i,i,ir) = 0 + T`? The answer is no, because `irdis_ws` is the supercell translations of `irvec`. The condition `irdist_ws(:,ideg,i,i,ir) = 0 + T` requires that the `irvec(:,ir)`, which locates in the the Wigner Seitz cell centered at the origin, to be `0`. 

Thus, if `irvec = 0`, then all the `irdist_ws` should also satisfy TB convention. If any `irdist_ws = 0`, then `irvec` must be `0`. 
In conclusion, the current `if` statement, that only test `irvec = 0`, is enough. 

Please point out my mistakes if I misunderstand anything.

Thanks,
Junfeng

paulatz:
Hello Junfeg,
I think you are making it too complicated. Note that if `i==j` and `ir==ir0` (i.e. R=0) than the degeneracy can only be 1, because an atom cannot be on the border of the WS cell centered on itself. In other words, if you set `r_m == r_n` and `R==0` than the minimal distance is for `T==0`, and it is zero. Also, note that `R+T` is always different from zero, except when `R==0` and `T==0`, because {R} are always inside the supercell, while {T} are always outside, and if `R'+T'==R+T` then `R'=R` and `T'=T`. I think you can prove this formally using the fact that the matrix of the (super-)cell vectors is invertible

qiaojunfeng:
Dear Lorenzo, 

Thanks for the comment. I forgot the condition `i == j` so things became complicated. When `i == j` and `R == 0`, the degeneracy is `1`, so `irvec` is identical to `irdist_ws`. In conclusion, it's safe to use either `irvec` or `irdist_ws` in the `if` statements. 

In terms of your previous comment on the symmetry of `wdist_ndeg`, I printed it out and found at some `ir` the `wdist_ndeg(:, :, ir)` were not symmetric. I will try to find out why.

Julen-ia:
Dear Junfeng and Lorenzo

Thank you for your replies. You obviously have more experience than me with this interpolation procedure.

From your discussion, I agree that the right thing to do is leaving the if statement as it is.

Thanks again

paulatz:
I've been thinking about wdist_ndeg being symmetric in i,j and I'm no more convinced that it is. 

I would say that if i_m+R is on the face (or edge, or vertex) of the super-WS cell of i_n, that i_n must be on the face (or edge, or vertex) of the super-WS of i_m+R. 

But we are not doing this, we are exchanging i_n-(i_m+R) with i_m-(i_n+R), which is not the same. There must however exist on R' that it is super-periodically equivalent to -R, and wdist_nedg(i,j,R) = wdist_ndeg(j,i,R'), which is a much weaker condition

qiaojunfeng:
Dear Lorenzo, 

Thanks! I agree with your arguments so `wdist_ndeg` is not necessarily symmetric. 
By the way, a little question about the code comment:
https://github.com/wannier-developers/wannier90/blob/6f61f72f6d7f10eb7f318c0d161e2253b3d3e95b/src/ws_distance.F90#L123-L134
Maybe it is better to write `function JW translated in the Wigner-Seitz around function IW`? 
This is more consistent with code and other comments:
https://github.com/wannier-developers/wannier90/blob/6f61f72f6d7f10eb7f318c0d161e2253b3d3e95b/src/ws_distance.F90#L33-L35
https://github.com/wannier-developers/wannier90/blob/6f61f72f6d7f10eb7f318c0d161e2253b3d3e95b/src/ws_distance.F90#L146-L148

paulatz:
Yes, you are right, also the general grammar and spelling of the comment is worse than my usual, I'll let you fix it if you want

qiaojunfeng:
Great, I think all the problems have been solved. 

I have regenerated the benchmark in `test-suite/tests/testpostw90_si_geninterp_wsdistance` and benchmarks in the SHC folders (using gcc 9.1.0, serial run), so the Travis-ci can pass now. 
By the way, I increased the tolerance a bit for the SHC tests, so issue #269 may pass as well.

giovannipizzi:
Thanks a lot @qiaojunfeng for spotting this bug and providing a fix, and all for the very useful discussions and bug scavenging :-)

We were discussing earlier today with @mostofi, @jryates and @ivosouza0 and we just had a doubt because of the relatively large change of the eigenvalues for the silicon interpolation with geninterp, so we decided to do additional tests before merging.

I just did them for Si (example 11). Indeed, the code before and after this PR returns the *same* band structure when interpolated with Wannier90, but the previous code (current develop, before these fixes) returned a *different* band structure between wannier90 and postw90 (with geninterp). 
With the changes discussed here, instead, also the geninterp interpolation goes back to being the same as the one of wannier90.

So, I think that the fix is correct and I'm going to merge it.

PR #284, opened 2019-07-29
Update kslice.F90

FreimuthFrank:
bnddataunit is not allocated --> leads to segmentation fault
Proposed solution: Add line
if(.not.allocated(bnddataunit))allocate(bnddataunit(num_wann))


src/postw90/kslice.F90
```diff
@@ -345,6 +345,7 @@ subroutine k_slice
 
         ! For gnuplot, using 'grid data' format
         if (.not. heatmap) then
+        if(.not.allocated(bnddataunit))allocate(bnddataunit(num_wann))
```
jryates: If you indent this line, the precommit hooks will pass.

giovannipizzi: Indeed, you should follow these short instructions: https://github.com/wannier-developers/wannier90/wiki/ContributorsGuide#automatic-pre-commits-spaces-indentation-

With these, you will have the automatic precommit hooks running and fixing indentation, spacing, ... - these are now required (in this case, there are a number of missing spaces on this line, e.g. after `.not.`, before `allocate`, ...

giovannipizzi: Hi @FreimuthFrank, do you think you can manage to fix the indentation as discussed above? If you do this, we can merge this PR. Thanks!

giovannipizzi:
Closed in favour of #301

PR #287, opened 2019-08-29
small fix for SLWF+C

hjunlee:
Dear all:

This commit does the following two things:

(1) enable the SLWF+C for all Wannier functions (WFs):
Currently, if the keyword of slwf_num is missing or slwf_num=num_wann is specified, selective_loc is set to .false. In this case, we have no way of performing the Wannierisation with centres fixed since the block of constraint is nested within the block of selective localisation.

I think that there might be some cases in which it is necessary to perform the SLWF+C for all WFs.

(2) very minor bug fix for printing the penalty function

Sincerely,

Hyungjun Lee


VVitale:
I believe there is a reason for this, which is the following: The sum of the centres of the Wannier function should be invariant modulo a lattice vector, therefore you cannot fix the centres of all Wannier functions to some target positions.
In the original paper by Marianetti et al. they never fix all centres but they only fix the centres of the selected WFs.

hjunlee:
You are right. I will close this commit.

PR #289, opened 2019-09-01
small restructuring of the routines for the AHC calculation with multiple Fermi energies and adaptive refinement

hjunlee:
Dear all:

This commit modifies the routines in berry.F90 so as to remove unnecessary multiple calls of berry_get_imf_klist in the berry module for calculations of AHC with adaptive refinement turned on and multiple Fermi energies (nfermi>1).
The current structure results in the duplication of some calculations unnecessarily.

With this commit, we can achieve much speed-up, in particular, in the case of large nfermi and rather dense adaptive-refinement k-mesh.

I made an effort to change the current structure of the berry module as little as possible even though I can achieve more speed-up with more restructuring.

Hyungjun Lee


giovannipizzi:
Don't worry, those tests are not critical.
Anyway, the reason is that more than 50% of the lines you added are untested.
Would you mind adding a test that goes through the new code path? Thanks!

giovannipizzi:
Hi @hjunlee - since we are planning to release v3.1 in about 1 week, if you would like to see this merged, it would be great if you could add the tests.
Do you think you would have the time to do it? 
[Here](https://github.com/wannier-developers/wannier90/tree/develop/test-suite/README.md) are the instructions if you never did it. If you have questions just let us know.
Thanks in advance!

hjunlee:
Dear Giovanni:

I added the test for calculation of anomalous Hall conductivity for bcc Fe with adaptive mesh refinement and Fermi energy scan.

Benchmark data comes from the calculation using the most recent develop version without my modification. Since my modification just removes unnecessary multiple calls, the results should be same and this is confirmed.

Sincerely,

Hyungjun Lee

giovannipizzi:
Thanks a lot!

PR #295, opened 2019-10-15
Minor fix and modification of compute_orb in pw2wannier90.f90

hjunlee:
Dear all:

This commit does the following three things:
(1) Fix the bug in the subroutine of compute_orb for the case with exclude_bands enabled.
(2) Modify the subroutine of compute_orb so as to avoid the unnecessary forward (inverse) FFTs.
(see https://github.com/wannier-developers/wannier90/pull/211)
(3) Additional minor modifications.

PS) Currently, compute_orb can't deal with the case with ultrasoft PPs and PAWs. However, even in the case using norm-conserving PPs, we have to maintain the storage for beta functions since non-local potential terms are calculated in h_psi.

Sincerely,

Hyungjun Lee


giovannipizzi:
Here is a description of the problem solved by this issue by @hjunlee, for future reference:

>  I attached the inputs and outputs for test of validation; I took example19 in the Wannier90 (W90) package (wannier90-develop/examples/example19) as an example.
> My modification just skips the step of inverse and forward FFT when periodic gauge condition is not necessary. Currently, W90 multiply e^{iG_0r} in real space when it is necessary to apply periodic gauge condition. For this, W90 performs both inverse and forward FFT; 
`psi(G) -> psi(r) -> psi'(r)=psi(r)*e^{iG_0r} -> psi'(G)`
> However, these FFTs are totally useless when G_0=0, that is, it is not necessary to apply periodic gauge condition. My modification addresses this problem and I already did similar things when mmn is calculated. (See #211. This is already merged)
> Consequently, as you can see the outputs of pw2wan (Fe.pw2wan.out), the walltime decreases from 22m to 16m. Theses time also include those for calculation of amn and mmn. Thus, the percentage of reduction in time only for calculation of uHu is large.
> Regarding the change in results:
> As you can see diff.before.after, there are only differences in beyond 13th floating points for Fe-morb.dat. I guess that the dummy operation of inverse and forward FFT makes tiny difference. But I believe that the result for keeping exactly wave function is more reasonable instead of performing useless inverse and forward FFT.

[PR295.tar.gz](https://github.com/wannier-developers/wannier90/files/4212290/PR295.tar.gz)

PR #301, opened 2019-12-04
Fix a missing allocation in kslice

giovannipizzi:
This was leading to a segfault as reported by Frank Freimuth

This commit replaces PR #284 that had tests failing (because of the pre-commit hooks)


giovannipizzi:
Closing this - in the meantime I had already removed the unused variable in #290 (here I just replicated the fix of the other issue, but in the meantime the variable got removed, so the problem does not exist anymore)

PR #307, opened 2020-01-23
add install target for makefile

### LINKED ISSUE #285: The project doesn't install anything
Please see how the Arch linux port installs the files: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wannier90

This is supposed to be done by your makefiles.

qiaojunfeng:
now can use `make install` to install the compiled binaries.
support `DESTDIR` and `PREFIX` variables.

By default, install
`wannier90.x` to `/usr/bin/wannier90.x`
`postw90.x` to `/usr/bin/postw90.x`
`w90chk2chk.x` to `/usr/bin/w90chk2chk.x`
`w90spn2spn.x` to `/usr/bin/w90spn2spn.x`
`w90pov` to `/usr/bin/w90pov`
`w90vdw.x` to `/usr/bin/w90vdw.x`
`libwannier.a` to `/usr/lib/libwannier.a`

Fixes #285


Makefile
```diff
@@ -8,6 +8,18 @@ TAR := $(shell if which gnutar 1>/dev/null 2> /dev/null; then echo gnutar; else
 
 default: wannier post
 
+PREFIX ?= /usr
+
+install:
+	install -d $(DESTDIR)$(PREFIX)/bin/
+	for x in wannier90.x postw90.x w90chk2chk.x w90spn2spn.x ; do \
+		if [ -f "$$x" ]; then install -Dm755 "$$x" "$(DESTDIR)$(PREFIX)/bin/$$x"; fi; \
```
giovannipizzi: I would remove the `-D` option, as this doesn't exist on the BSD version of `install` (e.g. on Mac OS X) but only on the GNU version (Linux). If I understand correctly it's just to create the folders, but you are already doing it in the line above? I would only leave `-m755`.

Makefile
```diff
@@ -8,6 +8,18 @@ TAR := $(shell if which gnutar 1>/dev/null 2> /dev/null; then echo gnutar; else
 
 default: wannier post
 
+PREFIX ?= /usr
+
+install:
```
giovannipizzi: Question (I don't know the answer): in other projects, does `install` depend on make, or it just installs what has been compiled? (Possibly it's ok as you did, just to double check)

giovannipizzi: Please add the `install` target to the `.PHONY` target at the end of the file - this ensures that `make install` works even if you create a file `install` in the current folder.

giovannipizzi (changes requested):
Thanks @qiaojunfeng !
Just one questions and two requests for change

Makefile
qiaojunfeng: Thanks! I didn't know this BSD difference, now `-D` are removed.

Makefile
qiaojunfeng: I checked the gnu make project, invoking `make install` will compile & install. So I add `default` target as a dependency of `install`.

qiaojunfeng: Has been added to `.PHONY`.

giovannipizzi (approved):
Thanks I think it's now ready to be merged!

PR #315, opened 2020-02-19
Fix the bug in the disentanglement step with SAWFs

### LINKED ISSUE #314: Possible bug in disentanglement when using SAWFs
Yusuke Nomura and I have discovered that the disentanglement procedure seem not to work as it should when imposing symmetry constraints as in SAWFs. We tested it on H3S (both in serial and in parallel) and while we obtain nice band interpolation using v2.1 this is not true with v3.0.0 (see attached figure). The disentanglement step in v3.0.0 seem to be stuck and the Delta (frac.) is not changing between iterations.

v3.0.0
 +---------------------------------------------------------------------+<-- DIS
  |  Iter     Omega_I(i-1)      Omega_I(i)      Delta (frac.)    Time   |<-- DIS
 +---------------------------------------------------------------------+<-- DIS

      1     236.28673237       6.89506462       3.327E+01      0.00    <-- DIS

        2     236.28673237       6.89506462       3.327E+01      1.01    <-- DIS

       3     236.28673237       6.89506462       3.327E+01      2.01    <-- DIS

       4     236.28673237       6.89506462       3.327E+01      3.02    <-- DIS

       5     236.28673237       6.89506462       3.327E+01      4.03    <-- DIS 

[H3S.pdf](https://github.com/wannier-developers/wannier90/files/4223857/H3S.pdf)

VVitale:
The bug was introduced in the following commit 5539b3c3610443e095bb93d56147a4abf0fd2bed

This should now fix it and also works for parallel execution.

Fixes #314


jryates:
There is an error in parallel which causes the test suite to fail. symmetrize_eps is not broadcast off the root node - hence if a none default value is set in the param file different nodes will apply the symmetry constraints to a different tolerance which propagates to give a different overall spread.

The fix is to add the following around line 6316 of parameters.f90
  call comms_bcast(symmetrize_eps ,1)

In addition the testsuite will fail on OSX because of the line is jobconfig 
 program = WANNIER90_DISENTANGLEMENT_SAWFs
If we replace the final s with capital S it is ok

PR #328, opened 2020-05-12
added writing ??_R files for wannier-berri

stepan-tsirkin:
Dear Wannier90 developers team, 

The present pull request adds some routines, which optionally  write the arrays AA_R, BB_R, CC_R, HH_R, SS_R and others evaluated within the get_oper.F90  to binary files.  This files are aimed to be be read by the wannier-berri python package, that I am developing.  
I also added to the manual the description of the 2 new variables, which  switch on this feature.
No other functionality of wannier90 or postw90 is affected. 

Best Regards,
Stepan

### DIFF
--- a/doc/tutorial/Makefile
+++ b/doc/tutorial/Makefile
@@ -1,4 +1,4 @@
-tutorial.pdf: tutorial.tex
+tutorial.pdf: tutorial.tex ../wannier90.bib
 	pdflatex tutorial.tex 
 	bibtex tutorial
 	pdflatex tutorial.tex
--- a/doc/user_guide/Makefile
+++ b/doc/user_guide/Makefile
@@ -1,6 +1,6 @@
 TEXFILES=$(wildcard *.tex)
 
-user_guide.pdf: $(TEXFILES)
+user_guide.pdf: $(TEXFILES) ../wannier90.bib
 	pdflatex user_guide 
 	bibtex user_guide
 	pdflatex user_guide
--- a/doc/user_guide/postw90params.tex
+++ b/doc/user_guide/postw90params.tex
@@ -154,6 +154,8 @@ magnetic moment per cell\\
 {\sc spn\_formatted}  & L & 
   Read a formatted {\tt seedname.spn} file\\
   {\sc berry\_curv\_unit} & S & Unit of Berry curvature\\ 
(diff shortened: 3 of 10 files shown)


stepan-tsirkin:
Dear All, 
within the current policy, is it obligatory to cover  the changes with tests? 
The new lines produce some binary files files, to be read by and external python code (wannier-berri) by means of FortranFile module.  
As far as binary files may be system/compiler  dependent, so it is not sensible to provide reference binary files, right?  
If I make a python script that parses those binary files and makes some control numbers from those, and the compares with the reference numbers,  will it work in the same way under different compilers? I am not sure how the testing farm works, when testing on different compilers? does it use a consistent python interpreter?

Can anyone instruct me on that?

Regards,
Stepan.

giovannipizzi:
Hi,
I think it's a good question.
In general, if we can add only tested code, this would be the best-case scenario, as the risk of introducing bugs (not much now, but even more in the future, if we need to refactor some of the functions) is huge. Of course there are cases were tests are just too complex to write, but if possible I'd try to think how to address this.

Now, in general indeed there is this issue of machine-dependent format of the fortran binary files. How big are these files, and would it be possible to export them as ASCII instead? I think this might be beneficial also for your code users in the long term, as it's hard to guarantee on an actual supercomputer that the python binaries are compiled with the same compilers as Wannier90 is.
If space is an issue, maybe you can gzip the files (after W90 generated them) and have your python library allow to read also gzipped files, for users who care more of space than on wall time?)

So, in a sense, I see it more of an issue of the chosen format, than just of the tests.
Anyway a simple way to test this, if you don't want to change format for now, is to just write a simple minimal test that reads the files in python, and we check on the build farm what happens.

I'm pinging here @sponce24 as he set up the farm and can give additional insights (or maybe even test out your PR on the test farm, once you put the tests, even before we merge it?)

stepan-tsirkin:
Thank you Giovanni,
of course for testing purposes those files may be kept small. 
I think in future I will proceed as follows:
1)I will add option to write ascii files. (and also to read them in my code)
2) make one test that compares with the reference ascii files.
3) another test that compares the writing of binary and ascii files. To make sure that neither is broken, and that the python compiles consistently. 

if (3) fails to work of test-farm, I will have avoid writing binaries.
I will update the PR later

stepan-tsirkin:
I have modified wannier-berri so that it reads wannier90 files directly ( chk, mmn,eig, uHu, spn) So this pull request is not needed  anymore.

Issue #329, opened 2020-05-15
RunTime error for hard fortran checks


zerothi:
https://github.com/wannier-developers/wannier90/blob/e77347fb6a70270e5dcbbc2f865b2be0566860d0/src/parameters.F90#L6384

This line will cause problems if compiled and runned with hard checks on limits.

Say if `gyrotropic_nfreq == 0`, then you'll have a problem. Typically this won't cause problems but it might. There are probably other cases (the line after ;)) where this may also happen.

stepan-tsirkin:
I think this problems cannot happen, because the following lines prevent it:
https://github.com/wannier-developers/wannier90/blob/e77347fb6a70270e5dcbbc2f865b2be0566860d0/src/parameters.F90#L1979

https://github.com/wannier-developers/wannier90/blob/e77347fb6a70270e5dcbbc2f865b2be0566860d0/src/parameters.F90#L1198

so, `gyrotropic_nfreq` and `gyrotropic_num_bands` are always positive.

zerothi:
I see, we have this problem here:

https://gitlab.com/siesta-project/siesta/-/merge_requests/7

I'll have to do some more digging to figure out exactly which version. But at least the error pops up... :(

stepan-tsirkin:
I see, it pops up in the library mode.
The gyrotropic things are used in the postw90, which is not used in the library mode. but the lines in parameters.F90 are executed anyway.

I do not see why this could happen, but I see that the neighboring lines have an if statement. probably there is a reason for that.

https://github.com/wannier-developers/wannier90/blob/e77347fb6a70270e5dcbbc2f865b2be0566860d0/src/parameters.F90#L6382-L6386

So it is safer to add analagous `if` statements to my gyrotropic lines. I fixed those two lines on the following branch: 
https://github.com/stepan-tsirkin/wannier90/tree/fix-gyrotropic-hard-checks

Please try and see if it solves your problem. Then I will make a pull request from it.

zerothi:
Thanks, I will forward and give it a go.

zerothi:
Thanks, this is now working.

Please merge! :)

jryates:
fixed by #335

PR #335, opened 2020-06-25
Broadcast variables for gyrotropic only when arrays are defined

stepan-tsirkin:
added checks before bcast for gyrotropic arrays 
As confirmed by Nick Papior ( @zerothi )  it fixes the issue #329


giovannipizzi (approved):
Thanks!

PR #338, opened 2020-07-15
Remove calls to pw90common_fourier_R_to_k

jaemolihm:
This PR removes all calls to `pw90common_fourier_R_to_k` in `postw90.x` and replaces them with `pw90common_fourier_R_to_k_new` or `pw90common_fourier_R_to_k_vec`.
I hope this correctly realizes what was intended by the comment "***REMOVE EVENTUALLY*** (replace with pw90common_fourier_R_to_k_new)".

Currently, `pw90common_fourier_R_to_k` is removed from the list of public routines, but its source code is not deleted. Maybe we can also delete the code itself?

I also updated the test-suite:
* added a new test for DOS with `spin_decomp = true`
* added a missing .gitignore for `testpostw90_fe_ahc_adaptandfermi`
* fixed a typo in `parse_dos_dat.py`.

Issue #296, opened 2019-10-24
Improving postw90 or rewriting it in python. 


stepan-tsirkin:
related to #280 , but regarding the postw90.x 

Dear all,

in this issue I would like to outline some problems of performance of postw90.x, and how I am advancing to solve them in another code, which I wrote (and continue developing) in python. And I welcome interested people to collaborate with me in this direction.

A number of modules of postw90.x  (e.g. berry.f90, gyrotropic.f90, boltzwann.f90) evaluate an integral over the Brillouin zone of some quantity.  The speed of such calculations are not perfect for several reasons:

1)  symmetries of the system are not accounted for, so summation runs over all k-points, instead of restricting only to the irreducible ones

2) A major part of computational time is used by  fourier transform. Using FFT would speed up the calculations.

3) Adaptive refinement, which proved useful for AHC, is implemented only there. And implementation in other quantities requires copy-pasting. 

4) Calculation of AHC for multiple fermi levels slow down calculations a lot.

5) Use of parameter use_ws_distance=true (which is default since v3.0) slows down calculation, because it is used at every call of Fourier transform. Instead, there is a way to re-define the real-space quantities only once, before looping over k-points.

To overcome these issues I started a new project in python [https://github.com/stepan-tsirkin/wannier19](url).  I called in Wannier19 by the year of starting.  It is now in active development stage, and not well documented. (I am the only developer so far). 
But already it can calculate Anomalus Hall conductivity and orbital magnetization very fast with high k-point density. In fact all 5 issues listed above are implemented there.  

It might sound surprizing (to someone)  that a python code is faster than a fortran code. But using of efficient libraries ( basically, numpy which includes fft) makes the "slowness" of python insufficient. And the use of flexibility and object-oriented model makes it easier to implement smarter algorithms. 

Implementation of the same algorithms in Fortran is possible, in principle, but not obvious to me.  However, if someone  is motivated to improve postw90, feel free to do it by analogy with my python code, I will provide explanations.

On the other hand, those who are interested in doing accurate calculations of AHC, or whatever quantity which is already implemented in postw90 (or even might be implemented), please contact me and join me in working on Wannier19.

I also wonder if the community is interested in eventual merging Wannier19 into Wannier90. Although they are in different languages, and communicate which each other only via input-output files, (wannierization is done by the good old Wannier90),   but it would be good to keep them consistent. Anyway, that's only when wannier19 is really mature enough. 

Thanks to all, who read this long post to the end!
Stepan.

[...]

Issue #275, opened 2019-05-31
Implementation of OpenMP parallelization of berry module


qiaojunfeng:
# What is new

This branch 
https://github.com/qiaojunfeng/wannier90/tree/shc_openmp
implements **OpenMP parallelization** of **berry module**, including spin Hall conductivity (SHC), anomalous Hall conductivity (AHC), orbital magnetization, Kubo optical conductivity, and shift current. 

# Why

It all started from an effort to improve the performance of calculating SHC. I am using a computer cluster with dozens of nodes, and each node contains 12 cores and 24GB memory, so each core can use 2GB memory at most. For large systems, a lot of Wannier functions are needed and `postw90.x` may stop running due to insufficient memory. The biggest obstacle is the real space matrics, e.g. `SR_R`, `SHR_R`, and `SH_R` in `get_oper.F90`. 

https://github.com/wannier-developers/wannier90/blob/280f38267410f085244732a1577f09433fa1500f/src/postw90/get_oper.F90#L31-L56

So in these cases, MPI parallelization leads to duplicated large real space matrics thus a waste of memory. OpenMP shared memory parallelization can be used to avoid memory waste. For example, 10 MPI process can be started on 10 nodes and each MPI process can start 12 threads, thus no waste of CPU cores and memory. 

# How to use

Two exemplary `make.inc` files are included: `config/make.inc.gfort.openmp` and `config/make.inc.ifort.openmp`. Two variables need to be modified for the compilation: `COMMS  = mpi openmp` and `FCOPTS = -O3 -fopenmp`. 

For running `postw90.x`, two environmental variables need to be set:

1. `OMP_NUM_THREADS=12`
2. `OMP_STACKSIZE=2G`

`OMP_NUM_THREADS` sets the number of threads to be started in each MPI process, usually the numbers of cores in each node.

`OMP_STACKSIZE` is particularly important for avoiding segmentation faults. It sets the stack size limit of OMP threads, and the default of `4MB` is too small for large calculations. I haven't tested the influence of the exact number on performance. On my cluster, the default value leads to segmentation fault and/or bus error, while `2G` works fine for me. 

For some systems, you may need to run `ulimit -s` to get rid of the stack size limit set by the operating system. 

Ref: https://stackoverflow.com/questions/13264274/why-segmentation-fault-is-happening-in-this-openmp-code

Then run `postw90.x` as usual. If you are using a job submission system, take care of the arguments for assigning the number of nodes and number of process to be used. 

[...]

PR #339, opened 2020-07-16
Optimization for use_ws_distance=true

jaemolihm:
Dear all,

This PR implements optimization of the use of Wigner-Seitz minimal distance replica selection (`use_ws_distance= true`) in the R to k Fourier transformation.

The optimization has been discussed by Stepan Tsirkin (#296, point 5) and is implemented in WannierBerri also by Stepan.
The WannierBerri website has detailed information on this optimization:
http://wannier-berri.org/methods.html#minimal-distance-replica-selection-method

Briefly, the real-space matrix elements are divided by the degeneracy factors `ndegen` and `wdist_ndeg` only once, in `get_oper` by the routine `operator_wigner_setup`. The `fourier_R_to_k` routines use a single set of R-point data: `nrpts_pw90`, `irvec_pw90`, and `crvec_pw90`, which is common, irrespective of `use_ws_distance`.
Hence, the optimized code is also much simpler than the previous one: the `fourier_R_to_k` routines do _not_ need two branches for `use_ws_distance=true` and `use_ws_distance=false`.

I tested the performance by running the calculation of testpostw90_fe_ahc, with berry_kmesh increased to 50*50*50. I used a single core. "Original" refers to the develop branch, commit 2f4aed6.
* `use_ws_distance=false`, original:  42.806 sec
* `use_ws_distance=true`, original: 136.720 sec
* `use_ws_distance=false`, optimized:  43.496 sec
* `use_ws_distance=true`, optimized:  43.237 sec

With the optimized code, using `use_ws_distance = true` requires no extra computational cost.

Also, I made the following additional modifications:
* Tighten tolerance of the shift current test. The range of the benchmark data in `testpostw90_gaas_sc_xyz` is 0 to 2E-5, but the previous absolute tolerance was 1E-4. I tightened the tolerance to 1E-6.
* Add four tests: `testpostw90_gaas_sc_xyz_scphase2`, `testpostw90_gaas_sc_xyz_scphase2_ws`, `testpostw90_gaas_sc_xyz_wsdistance`, `testpostw90_fe_kpathmorbcurv_ws`. Now, all the fourier_R_to_k routines are covered by tests.
* Add clocks (activated when `timing_level>2`) to some fourier_R_to_k routines
* Move call to `ws_translate_dist` from `pw90common_fourier_R_to_k` to `pw90common_wanint_setup` (This is also done in #275: Technical details, point 2)
* Remove source code of `pw90common_fourier_R_to_k` (the calls are removed in #338)

This PR is written assuming that #338 is merged.

Best regards,
Jae-Mo


src/postw90/gyrotropic.F90
```diff
@@ -484,7 +484,7 @@ subroutine gyrotropic_get_k_list(kpt, kweight, &
       pw90common_fourier_R_to_k_vec
     use w90_wan_ham, only: wham_get_eig_deleig, wham_get_D_h
 
-    use w90_get_oper, only: HH_R, SS_R, AA_R
+    use w90_get_oper, only: HH_R, AA_R
```
stepan-tsirkin: Actually, HH_R is also not needed here

stepan-tsirkin (commented):
Dear Jae-Mo,

thanks for implementing this in wannier90. I have one quick question - is the writing of "_hr.dat" and "_tb.dat" files affected by this pull request? I think it would be very useful to have already re-defined HH_R and AA_R  in those files.  This may be done in a separate pull request though.

Stepan.

jaemolihm:
Dear Stepan,

No, the writing of "_\*.dat" files is not affected by this pull request.
 I only modified the source codes of `postw90.x` (those inside the `src/postw90/` directory), while the _*.dat files are written by `wannier90.x`. 

Jae-Mo

src/postw90/gyrotropic.F90
jaemolihm: I removed it: 93ac701. ( I also moved `pw90common_fourier_R_to_k_vec` outside the loop: my mistake with changing from `fourier_new` to `fourier_vec`.)

There is a separate branch where I removed (almost) all of the unused variables and USE statements in `postw90.x` using the `-check all -warn all` flag of ifort compiler, but I thought it would be better to have a separate PR for that.
See https://github.com/jaemolihm/wannier90/compare/use_ws_opt...jaemolihm:wigner_opt.

jaemolihm:
I also should note that the optimization may not work with `effective_model = true .and. use_ws_distance = true`. The reason is the setup for ws_distance is called in `pw90common_wanint_setup`, before the R vectors are read (which is done in `get_HH_R`).
It seems that  `effective_model = true` is intended not to work with `use_ws_distance = true`. If so, it would be better to have a check in parameters.F90.

I think `effective_model = true .and. use_ws_distance = false` case should work fine, but I have not tested it, as I could not find a working example.

stepan-tsirkin:
I agree, there is no need to combine `effective_model = true .and. use_ws_distance = true.` . 

`use_ws_distance = true.` aims to correct the R-space matrix elements when converting from the abinitio k-grid

As I understand - effective model reads matrix elements  which are already in R-space (e.g. a handmade TB model). So there maybe no any "abinitio k-grid" implied. 

You might add a parameter check to avoid this situation. 
`effective_model`  is not documented , and hard to say exactly how it should work. Looks like the format of the `_HH_R.dat` file is similar to that of `_hr.dat`, except that degeneracies are not written. You may try to  generate `_hr.dat`, remove degeneracies and rename it to `_HH_R.dat` , and try to run this way. But it would be good to find someone who implemented and/or use `effective_model=.true.`   @ivosouza0 ,  were you involved in that?

jaemolihm:
I will reorganize changes made in this PR and in #338, and reopen separate PRs for 1) improved postw90 tests, 2) removing `pw90common_fourier_R_to_k`, and 3) Wigner-Seitz optimization.

I will also update to include already merged PRs, including #349.

PR #340, opened 2020-08-03
Very minor issue in the subroutine of internal_cube_format in plot.F90

hjunlee:
Dear all:

There were very minor typos(?) in the subroutine of internal_cube_format in plot.F90:
ngx, ngy, and ngz correspond to dffts%nr1, dffts%nr2, and dffts%nr3, respectively. That is, ngx is the number of grids along the first direct lattice vector, etc.

Sincerely,

Hyungjun Lee


mostofi (approved):
Thanks for spotting this - your update looks correct to me.

PR #342, opened 2020-09-02
Modification needed for SAWF in Wannier library mode.

sponce24:
The .dmn file needs to be read.


hjunlee:
Hello, Samuel:

Regarding SLWF, please see the comment at https://github.com/wannier-developers/wannier90/pull/287#issuecomment-526352354 .
Regarding SAWF, if you have in mind the use of SAWF in EPW, I think that dmn files should be calculated in EPW in order to make sure the gauge consistency.

Sincerely,

Hyungjun Lee

sponce24:
Hello Hyungjun, 

For SLWF, it seems to work to constrain all centers (I tested it). The reason that they never constrained all centers is due to their DEFT-based application. Therefore it think it would not hurt to allow it (potentially with a Warning message). 

For SAWF, the fix is needed for any code using w90 in library mode. 
I also tried using it with EPW indeed. 
It also seems to work but I agree it would be better to compute it internally, but this is a different question that we can solve in the EPW repo. 

Best, 
Samuel

src/parameters.F90
```diff
@@ -827,7 +827,7 @@ subroutine param_read()
       if (slwf_num .gt. num_wann .or. slwf_num .lt. 1) then
         call io_error('Error: slwf_num must be an integer between 1 and num_wann')
       end if
-      if (slwf_num .lt. num_wann) selective_loc = .true.
+      if (slwf_num .lt. num_wann + 1) selective_loc = .true.
```
giovannipizzi, line 834: Hi @sponce24 can you explain this change? I think `selective_loc` is meant to be set if only some of the WFs are selectively localised? @VVitale maybe can also comment

jryates:
Samuel - apologies for the delay in looking at this PR.

Could you let us know if you would still like this merged (and maybe answer Giovanni's question). If so we will probably merge when we have a coding meeting on the 4th August.

src/parameters.F90
VVitale: Hi @sponce24 @giovannipizzi and @hjunlee, apologies for the very late reply.  As I mentioned in #287, you should not be able to arbitrarly fix "all" centers of the Wannier functions and expect this to work for an isolate manifold with number of bands = number of WFs, since the vectorial sum of the centers is a gauge invariant quantity (modulo a lattice vector). Whether this is possible for entangled bands, I am not really sure...

jryates:
Superseded by later developments.

PR #345, opened 2020-09-28
fix typo in documentation of wannier_plot_spinor_mode parameter

qiaojunfeng:
There are 2 typos in the doc of section 2.9.8 `wannier_plot_spinor_mode`, should be 
- up. |φ| × sign(Re{φ}) if wannier_plot_spinor_**phase** = true, otherwise |φ|

instead of 
- up. |φ| × sign(Re{φ}) if wannier_plot_spinor_**mode** = true, otherwise |φ|

According to code:
https://github.com/wannier-developers/wannier90/blob/10e54f32fc2ca7ddfcffc2eb34ae1fc45a86da2d/src/plot.F90#L1169-L1181


jryates (approved):

jryates:
Thanks - good spot!
 I've approved this, but something has got stuck with our CI (the tests passed fine). So will merge when this is resolved.

PR #349, opened 2020-11-20
implement GitHub Actions for all tests, replacing Travis

jimustafa:
- add the `pre-commit` action (like [this](https://github.com/pre-commit/action))
- essentially revert 9036cda245d2441ffd8a562d2b21681b0085a4b5


giovannipizzi:
Thanks @jimustafa !
Since you are switching to GitHub actions (great!) would you have the time to also switch completely, creating another step for the standard tests?

jimustafa:
Hey @giovannipizzi. Looks like this PR may have successfully implemented the migration to GitHub Actions. Not sure that the workflow will run on the PR... The changes may need to be merged into `develop`, and then the workflow should run on the merge commit. Everything appears to be working on my branch [here](https://github.com/jimustafa/wannier90/actions/runs/462819981). If there is a problem, please revert, and we can try again. Thanks!

giovannipizzi:
Thanks! I'm going to use admin rights to merge even if the old travis tests are not running, hoping that all works, and will then re-activate branch protection for github actions.
The parallel tests are slower than on travis, oddly (seems 18 minutes vs 13 minutes). Maybe we can switch to using 2 CPUs instead of one, or maybe finetuning the compilation.

giovannipizzi:
Seems to be working! Pinging @jryates @mostofi @VVitale @ivosouza0 for info - current pull requests need to be updated to include these commits before merging, so tests can run and pass. We do not depend anymore on travis.

PR #350, opened 2020-12-20
A new option to SHC

manxkim:
This PR implements Ji Hoon Ryoo's spin Hall conductivity using Wannier interpolation, tested and publised(Phys. Rev. B 99, 235113). It is slightly different from Junfeng Qiao's (Phys. Rev. B 98, 214402), which uses some approximations.

To calculate SHC with Ryoo's method, add the parameter "shc_ryoo=.true." in postw90.x input description. It can be used together with other shc parameters by Qiao. Also, it requires .sHu and .sIu files generated from pw2wannier90.x of quantum espresso, implemented by Ryoo and me.

Regards,
Minsu Ghim
Seoul National University


jryates:
Thanks for this.

I haven't read the newer paper, or thought about the physics yet. But a few comments:

1- I think it would be more in keeping with W90 style to have a parameter shc_method with values qiao or rhoo (or perhaps something more physical?) rather than shc_ryoo=.true. [perhaps @giovannipizzi or @mostofi can comment]

2- Can you add a test case?

3- Can you update the user guide (after the keyword is decided)

4- There are a few debug lines commented out. It is cleaner for the code base if these are removed. We do ignore this sometime if there is good reason to leave in uncommented code - so use your judgement.

5- The pre-commit checks failed. You need to install the pre-commit hooks, see
https://github.com/wannier-developers/wannier90/wiki/ContributorsGuide

giovannipizzi:
Hi all!
1. I agree, it would be better to have something like `shc_method` or `shc_implementation`
2. Indeed, is there an easy way to summarise the difference between the two methods, ideally in one word? :-) Otherwise I guess we're left to the first-author's name
3. I suggest that @qiaojunfeng helps reviewing this PR, as he's the author of the "other" method :-)
4. about the comment `it requires .sHu and .sIu files generated from pw2wannier90.x of quantum espresso, implemented by Ryoo and me`, did you already commit this into the [`pw2wannier90.x` of the Quantum ESPRESSO code](https://gitlab.com/QEF/q-e/-/blob/develop/PP/src/pw2wannier90.f90)? If not, would it be possible to make this available before we merge this PR? Ideally committing directly there, but we also keep a copy of the "patched" versions here, in the pwscf folder (i.e. the file that needs to be replaced for the most recent QE release to work - 6.7 at the moment).

manxkim:
Thank you all for comments,

Because I made too many mistakes of pushes and commits("This branch is 740 commits ahead of wannier-developers:develop."), I decided to move to a new branch and pull request newly. My apologies for that, please see https://github.com/wannier-developers/wannier90/pull/353

PR #355, opened 2021-02-11
simplified projectability vs e table

ahzeeshan:
The current code produces a `p_vs_e.dat` file with 231000 rows but it should have 21,000 rows corresponding to 21 bands times 1000 k-points. Modified the code to do that. The error seems to be due to looping over the number of bands and writing to file at the end of each iteration. Please correct me if that was the intention.

### DIFF
--- a/examples/example32/generate_weights.sh
+++ b/examples/example32/generate_weights.sh
@@ -13,17 +13,9 @@ rm -f p_vs_e.dat
 [[ -f "proj.out" ]] || { echo "proj.out not found!"; echo "Aborting!"; exit 1; }
 
 # Get energies and projectability in the correct order
-for i in {1..21}; do
-        echo $i
-        if [ $i -lt 10 ]; then
-           line=`echo "e(   $i)"`;
-        else
-           line=`echo "e(  $i)"`;
-        fi;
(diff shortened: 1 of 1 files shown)


VVitale:
You are right the code generates 23100 lines, where there should be 21000. This is because the paste command is inside the loop when it should be outside. However, your proposed solution changes the order of projectability vs energies, in the sense that now the outer loop is on k-points, whereas the original outer loop was over bands. I am going restore the original ordering with the bugfix.

VVitale (approved):

ahzeeshan:
@VVitale Does the ordering matter? At the end, the user has to (scatter) plot p vs e and perform fitting to obtain the two parameters mu and sigma. I have checked the fitting using `scipy.curve_fit` and I get the same values of mu and sigma for W as the paper. The advantage of the code I proposed is it can be used for any number of bands and `grep -A4` in the current code is making strong assumptions on the `proj.out` file, namely that the projectability is printed within the next four lines, and ``echo "e(  $i)"`;` assumes some way of printing of energies which has to be again modified when you have >100 bands. The final p_vs_e.dat file can always be reshaped when doing the fitting if one wants them band ordered, for example, to exclude some bands.

VVitale:
@ahzeeshan That's true. I have restored your version.

PR #362, opened 2021-03-20
run workflows explicitly on Ubuntu 20.04 with `numprocs=2`

### LINKED ISSUE #361: GitHub Actions workflow failing on Ubuntu-20.04 runner
First noticed in (but not a consequence of) 513694b, the test suite is failing for the parallel build. This appears to be due to GitHub changing `ubuntu-latest` from 18.04 to 20.04 (compare [this](https://github.com/wannier-developers/wannier90/runs/2054248431) to [that](https://github.com/wannier-developers/wannier90/runs/2121414288)).

A quick fix would be to configure the workflow to explicitly run on `ubuntu-18.04`. I will open a PR to this effect.

The issue is likely different package versions in Ubuntu 20.04. Although there does not appear to be any compilation problems, the parallel executables may be broken.

jimustafa:
attempts to close #361

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,3 +59,11 @@ jobs:
           ./.github/workflows/run_tests.sh
       - name: compute coverage
         uses: codecov/codecov-action@v1
+      - name: archive test results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-results_parallel=${{ matrix.w90-binary-parallel }}
+          path: |
+            test-suite/tests/test*/test.err*
+            test-suite/tests/test*/test.out*
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,14 +10,14 @@ on:
 
 jobs:
   pre-commit:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


giovannipizzi:
Hi Jamal, thanks!
Apparently, rather than freezing to 18.04, it seems that there is a better fix.
In particular Jerome Jackson reported that 
> after a few tests we tracked down the problem to openmpi (version 4)'s reluctance to run more mpi tasks than there are slots available: for Github actions the limit is (I think) 2.  Allowing oversubscription requires a openmpi specific option to mpirun (so intel's mpi system then fails); instead we elected to run parallel tests only with two mpi tasks.  Presumably this problem has appeared now because the default version of openmpi used by ubuntu-latest has been recently increased.

This fix has already been added to #360, but since that is a bigger PR and might take longer to fix, maybe you want to do that fix here (use `./run_tests --category=default --numprocs=2` in `.github/workflows/run_tests.sh`) so that, if this works, this fix can already be merged?

jimustafa:
Nice catch by Jerome Jackson! Yes, the proposed changes appear to work. Thanks!

giovannipizzi (approved):
Thanks!

PR #352, opened 2021-01-08
Improving test-suite

jaemolihm:
This PR improves the test-suite by making the following changes:

* Tighten tolerance of the shift current test. The range of the benchmark data in testpostw90_gaas_sc_xyz is 0 to 2E-5, but the previous absolute tolerance was 1E-4. I tightened the tolerance to 1E-6.
* Add five new tests: `testpostw90_gaas_sc_xyz_scphase2`, `testpostw90_gaas_sc_xyz_scphase2_ws`, `testpostw90_gaas_sc_xyz_ws`, `testpostw90_fe_kpathmorbcurv_ws`, `testpostw90_fe_dos_spin`. Now, all the `fourier_R_to_k` routines are covered by tests.
* Add a missing `.gitignore` for `testpostw90_fe_ahc_adaptandfermi`
* Fix a typo in `parse_dos_dat.py`.

The running time of the tests do not increase much. When I ran all the tests on my laptop, I got almost the same time.
Current develop branch: 5m 32s
This PR: 5m 34s

(If this time becomes too long, I think we should make some "bottleneck" tests cheaper. For example, `testpostw90_fe_ahc_adaptandfermi` and `testpostw90_pt_shc` took 72 and 74 seconds, respectively.)

This PR implements a part of the closed PRs #338 and #339.

PR #368, opened 2021-06-02
Correction for shift current with finite eta

jaemolihm:
This PR implements a correction term for the shift current that should be added to make the result independent of the phase convention. See [Jae-Mo Lihm, "Comment on "Ab initio calculation of the shift photocurrent by Wannier interpolation", Phys. Rev. B 103, 247101 (2021)](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.103.247101)  for the derivation and further analysis.

### List of changes
* Implement the correction term. Add `sc_use_eta_corr` input parameter for choosing to use it or not. Default is set to true (using correction).
* Fix a bug in shift current implementation regarding the use of sc_eta (also explained in the Comment linked above).
* Changed shift current test: Add `sc_use_eta_corr = .false.` (do not use the correction), update reference data to reflect the bugfix.
* Update user guide.
* Add tests set for `sc_use_eta_corr = .true.` (`testpostw90_gaas_sc_eta_corr`)

Note that currently, the tolerance of shift current tests are too loose, so the tests are not very meaningful. (For example, it does not raise error even if the value of `sc_use_eta_corr` is changed, because the change in the output data is smaller than the test tolerance.) #352 is an open PR to tighten the tolerance.


Julen-ia:
The code correctly incorporates the correction formula proposed in the Comment to the original ab-initio shiftcurrent PRB, needed when a finite broadening is used

ivosouza0 (approved):

PR #370, opened 2021-06-19
plot_wannier is parallelized

KensukeKurita:
Currently, it takes a very long time to plot the Wannier function.
This is because all k-points are calculated serially in plot_wannier() even if you use mpi.

So, I parallelized plot_wannier().
The change is simple:
- assign the computation at each k-point to multiple cores.
 `do loop_kpt = displs(my_node_id) + 1, displs(my_node_id) + counts(my_node_id)`

However, I did not change any functions in plot_main() except plot_wannier().

In my environment, when I set the number of parallelism to 10, **the time taken for plotting is about one tenth**.


src/Makefile.2
```diff
@@ -109,7 +109,7 @@ disentangle.o: ../disentangle.F90 constants.o io.o parameters.o sitesym.o comms.
 wannierise.o: ../wannierise.F90 hamiltonian.o constants.o io.o utility.o parameters.o sitesym.o comms.o
 	 $(COMPILER) $(POSTOPTS) $(FCOPTS) -c ../wannierise.F90
 
-plot.o: ../plot.F90 constants.o io.o utility.o parameters.o hamiltonian.o ws_distance.o
+plot.o: ../plot.F90 constants.o io.o utility.o parameters.o hamiltonian.o ws_distance.o comms.o:w
```
qiaojunfeng: there is an additional `:w`

qiaojunfeng (changes requested):
Thanks! 
Tested with both non-spinor and spinor modes, works well.
There's only one typo that needs to be fixed in `src/Makefile.2`.

src/Makefile.2
giovannipizzi: Indeed, this makes all tests fail to compile - great if this could be fixed and then we can merge the PR

KensukeKurita: I appreciate your review.
I fixed the typo in `src/Makefile.2` and did recommit.

giovannipizzi (approved):
Thanks!

PR #397, opened 2022-02-23
add comment delimiter to first line comment in NNKP file

jimustafa:
Given that the NNKP file has essentially the format as the WIN file, maybe it is best if the comment on the first line is delimited by `#` or `!`.


jryates:
Seems reasonable - but I'll hold off approving until I've discussed with @giovannipizzi  and @mostofi

giovannipizzi:
I don't see an immediate problem with this - that line should be a comment anyway so it should be essentially ignored by any code, right? So adding a # in front shouldn't break and seems a good idea. So I'm OK

jryates (approved):

Issue #398, opened 2022-02-24
Several examples have an errant `kpoints` block


jimustafa:
There are several of the `wannier90` examples that have WIN files with an errant `kpoints` block; each line in the block has 4 numbers, instead of 3, as specified in the user guide. It appears that the `kpoints` were generated with the weights as the 4th column. While this is inconsequential, as `wannier90` appears to only read the first 3 numbers in each line, it probably should be fixed so that the examples are consistent with the user guide. If you all agree, I will submit a PR.

jryates:
Well spotted. Yes, that would be good. Thanks!

jryates:
fixed by #401

PR #401, opened 2022-02-25
fix some examples having an errant `kpoints` block

jimustafa:
- a `kpoints` block should only have 3 columns
- removed the 4th column in `kpoints` block of some exmaples


jimustafa:
Attempts to close #398.

The large `kpoints` block in some of the examples brings two things to mind... Are that many `kpoints` necessary for those examples? Is it possible to just use the `mp_grid` parameter to generate the `kpoints` for the NNKP file and internally? Having both the `mp_grid` and `kpoints` seems redundant, but I am probably missing an important use case...

jryates (approved):

jryates:
To address your questions:
 1- Probably not. (although as long as the test suite runs in a reasonable time, it is not the worst thing)
 2- I captured this in the new issue #403

PR #402, opened 2022-02-25
trim trailing semicolons

### LINKED ISSUE #400: Purge code base of trailing semicolons
As trailing semicolons are unnecessary, it may be best to purge them from the code base. They also seem to cause problems with `fprettify` (see pseewald/fprettify#96). There has been a feature request (pseewald/fprettify#120) to allow for removing trailing semicolons, but it appears that the `fprettify` project may not be maintained anymore.

Going further, I would suggest multiple statements on a line should also be avoided.

jimustafa:
Attempts to close #400.

### DIFF
--- a/src/postw90/berry.F90
+++ b/src/postw90/berry.F90
@@ -423,7 +423,7 @@ contains
                     num_valence_bands, effective_model, have_disentangled, seedname, stdout, &
                     timer, error, comm)
       if (allocated(error)) return
-      kdotp_nbands = size(pw90_berry%kdotp_bands); 
+      kdotp_nbands = size(pw90_berry%kdotp_bands)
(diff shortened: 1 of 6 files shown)


jimustafa:
Should we also go through and break up lines with multiple statements. While taking more vertical space, it may improve readability.

Side question... Is there an issue with replacing
```
lmn(1, counter) = l; lmn(2, counter) = m; lmn(3, counter) = n
```
with
```
lmn(1:3, counter) = (/ l, m, n /)
```

Maybe a small performance hit if it leads to creating a temporary array? Does not seem to cause warnings about array temporaries in a small test with `-fcheck=all`.

jryates (approved):

jimustafa:
Please advise on how to resolve the conflicts. Thanks.

PR #406, opened 2022-03-16
Update Makefile to install a package config file.

pkestene:
This PR provides minimal modifications to create and install a pkg-config file so that users of wannier library can more easily detect wannier from a cmake or an autotool build system.

If a user installs wannier, and then setup env variable like this 
```shell
export PKG_CONFIG_PATH=$WANNIER_ROOT/lib/pkgconfig
```

Then wannier library can be integrated into a cmake build like this:
```cmake
find_package(PkgConfig REQUIRED)
pkg_check_modules(WANNIER QUIET IMPORTED_TARGET wannier)
if(WANNIER_FOUND)
    add_library(wannier::wannier ALIAS PkgConfig::WANNIER)
endif()
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,13 @@ default: wannier post
 
 PREFIX ?= /usr
 
+VERSION_MAJOR = 3
+VERSION_MINOR = 1
+VERSION_PATCH = 0
+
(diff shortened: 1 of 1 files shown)


jryates:
The pre-commit fails due to issue #407 no problem with this PR.

giovannipizzi (approved):

PR #337, opened 2020-07-12
fix the issue in plot_bvec

hjunlee:
Dear all:

This commit fixes the issue in writing *.bvec file which is necessary for the evaluation of velocity matrix elements in EPW.
Previously, the writing of bvec file is triggered by the input flags which are not directly related to it.
Additionally, I removed one of two open statements and increased the number of significant digits in *.bvec file so that it matches those from kmesh.pl .

On behalf of the EPW developers' team,
Hyungjun Lee

PR #409, opened 2022-03-29
Increase digits printed by plot_bvec

jaemolihm:
`seedname.bvec` file is used by EPW, and I found small but nonzero (~1E-6) relative error from the round-off errors in bvec. Such a small error rarely affects physics but it makes debugging tricky. So I believe it is worth changing.

In addition, `testw90_bvec` did not test the bvec data (it only checked the sizes). I updated the parser to check bvec data and updated the reference data.


hjunlee:
@jaemolihm I already changed the number of digits here (https://github.com/wannier-developers/wannier90/pull/337) and for the moment, I don't want further changes.

In addition, this part is only related to EPW and I would like to ask all people who want to do MRs related to EPW in Wannier90 to directly contact me, the lead developer of EPW, before doing MRs.

jaemolihm:
@hjunlee This PR is a change to Wannier90 source code, not EPW source code. This code is in the Wannier90 upstream. Thus, I believe the code should be managed in the same way as other parts of Wannier90, i.e. as an openly-developed community code. I have followed the Wannier90 contributors guide and opened a Pull Requests.

(I think the situation is different from the case of EPW in QE, where the whole EPW/ folder is managed by the EPW team. For changes to the EPW source code, I will certainly follow your instruction.)

In addition, it is not correct that this part (subroutine plot_bvec) is related only to EPW. One can write the bvec file using the write_bvec flag, whether or not one is using EPW. I have faced problems using the seedname.bvec file for my own program (not EPW), thus this PR.

If the reason for objection is that such a change in printed digits is only a quick-and-dirty fix and one should aim for a more fundamental change (e.g. binary IO or directly accessing the variable via library mode), I would accept the argument and close this PR.

jaemolihm:
@hjunlee In the future, I will make sure I contact and notify you when I make a PR that would impact the EPW code.

giovannipizzi (approved):
Thanks @jaemolihm! This change is backward compatible and should not affect/break the behaviour of external codes, and I think it's an important improvement (and also comes with tests - thanks!), so I am going to merge this. 
If this creates problems, we can discuss in a new issue.

PR #415, opened 2022-04-14
Update wan_ham.F90

Emad-Rezaei:
Adding subroutines to compute the derivative of velocity with respect to wave vector and the Omega operator.


jryates:

PR #419, opened 2022-04-14
Update readwrite.F90

Emad-Rezaei:
Reading and writing the variables for the nerwann module


jryates:

Issue #461, opened 2024-01-01
Issue with Parallel Compilation of Wannier90.


hongyi-zhao:
Hi here,

I have encountered an issue while trying to compile Wannier90 in parallel, as shown below:

```shell
# The OS information:
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy

# The toolchain use here is Intel oneAPI:
$ module load compiler mkl mpi
Loading compiler version 2021.1.1
Loading tbb version 2021.1.1
Loading debugger version 10.0.0
Loading compiler-rt version 2021.1.1
Loading dpl version 2021.1.1
Loading mkl version 2021.1.1
Removing compiler-rt version 2021.1.1
Loading compiler-rt version 2021.1.1
Removing tbb version 2021.1.1
Loading tbb version 2021.1.1
Loading mpi version 2021.1.1

# The problems:
werner@X10DAi:~/Public/repo/github.com/wannier-developers/wannier90.git$ make all -j44
(cd ./utility/w90pov && make )
(cd ./utility/w90vdw && make )
(cd ./src/obj && make -f ../Makefile.2 wannier)
(cd ./src/obj && make -f ../Makefile.2 libs)
make[1]: Entering directory '/home/werner/Public/repo/github.com/wannier-developers/wannier90.git/utility/w90pov'
(cd ./src/obj && make -f ../Makefile.2 post)
ifort -O2 -c src/general.f90 -o obj/general.o
make[1]: Entering directory '/home/werner/Public/repo/github.com/wannier-developers/wannier90.git/utility/w90vdw'
ifort -O2 w90vdw.f90 -o w90vdw.x
(cd ./src/obj && make -f ../Makefile.2 w90chk2chk)
(cd ./src/obj && make -f ../Makefile.2 w90spn2spn)
make[1]: Entering directory '/home/werner/Public/repo/github.com/wannier-developers/wannier90.git/src/obj'
mpiifort -DMPI -DMPI90 -O2 -c ../constants.F90
make[1]: Entering directory '/home/werner/Public/repo/github.com/wannier-developers/wannier90.git/src/obj'
mpiifort -DMPI -DMPI90 -O2 -c ../constants.F90
mpiifort -DMPI -DMPI90 -O2 -c ../error_base.F90
mpiifort -DMPI -DMPI90 -O2 -c ../error_base.F90
[...]

PR #472, opened 2024-02-16
Fix the issue of parallel compilation of Wannier90

hongyi-zhao:
Fix the issue of parallel compilation of Wannier90 as discussed here:
https://github.com/wannier-developers/wannier90/issues/461#issuecomment-1946675452


giovannipizzi:
Already fixed in #473

PR #474, opened 2024-02-16
Call mpi_finalize in the end of postw90 dry run

### LINKED ISSUE #346: dryrun do not terminate mpi properly
Hi, developers

I found a small problem in the postw90 code:

In the postw90.f90, with dryrun==true, the program do not finalize mpi, it can be easily solved by inserting "call COMMS_END" before the stop statement.
 
  if (dryrun) then
    if (on_root) then
      write (stdout, *) ' '
      write (stdout, *) '                       ==============================='
      write (stdout, *) '                                   DRYRUN             '
      write (stdout, *) '                       No problems found with win file'
      write (stdout, *) '                       ==============================='
    endif
    stop   <----
  endif

I believe this only affect dryrun in parallel mode but could be a bit confusing in some case. If would be nice if you fix it.

Best wishes
Wenhao

npaulish:
This fixes #346 to avoid incorrect termination when dryrun is done with MPI.

### DIFF
--- a/src/postw90/postw90.F90
+++ b/src/postw90/postw90.F90
@@ -330,7 +330,12 @@ program postw90
       write (stdout, *) '                                   DRYRUN             '
       write (stdout, *) '                       No problems found with win file'
       write (stdout, *) '                       ==============================='
+      close (stdout)
+      close (stderr, status='delete') ! this should not be unit 0
     endif
+#ifdef MPI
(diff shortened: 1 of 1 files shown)


giovannipizzi (changes requested):
Remember to close the stdout and stderr units

giovannipizzi (approved):

PR #486, opened 2024-02-19
Add mpi to `bands_plot`

### LINKED ISSUE #452: Parallelization of band structure calculation
Dear all，

       The band structure calculation seems not parallelized, which is slow, especially for large system.

       Is it possible that someone make it faster in next version?

I would be grateful for your help.

jiang-yuha0:
Resolve #452, enable parallel run for `plot_interpolate_bands` in `src/plot.F90`.
Add `io_stopwatch` for the main kpoints loop in `plot_interpolate_bands`.


mywangbit:
Thanks for changing. I also wonder know that when the new version of wannier90 is online?

JeromeCCP9 (approved):
Straightforward parallelisation of the band plotting, together with test case.

PR #490, opened 2024-02-23
Fix bounds error when assigning M matrices

qiaojunfeng:
With gfortran `-fcheck=bounds` the chkpt distribution raise a runtime error:

```
Array bound mismatch for dimension 1 of array 'm_matrix_local' (12/8)
```

This fixs test failure of #486.


JeromeCCP9 (approved):
Yes, this is absolutely correct.  Bug was introduced (by myself) when reorganising m_matrix array use.

PR #500, opened 2024-04-28
Fixed bug in writing checkpoint file in parallel mode

sjhong6230:
While computing various quantities using `wannier90.x` and `WannierBerri`, I reported the following issue.

https://github.com/wannier-berri/wannier-berri/issues/325#issue-2262487002

With more tests, I found out this also happens for `postw90.x`. Also, it happens only for parallel executions.

When I inspected the code, I noticed a bug in the `write_chkpt` subroutine in the `wannier_prog.f90` file. 
Since `common_data%u_matrix` and `common_data%u_opt` are global, they should not be reduced by MPI routines.

Fixing this point gives consistent results with Wannier90-3.1.0. I also suspect the reason for the above issue is the same.

The reason that this bug was not reported in tests is that the checkpoint files were generated from previous versions.

This bug needs an immediate fix and this PR should be considered urgent.


sjhong6230:
Comment on test-suite: To prevent such bugs, I recommend adding test sets that uses `wannier90.x` and `postw90.x` consecutively.

JeromeCCP9 (approved):
This bug fix is correct.

JeromeCCP9:
Much appreciated, Seung-Ju!  J.

PR #498, opened 2024-04-22
Implementation of Stengel-Spaldin Functional in Wannierization

sjhong6230:
The main differences for this pull request are shown below.

- Implemented the Stengel-Spaldin functional [1] in Wannierization.

- Added `use_ss_functional` parameter. Uses Stengel-Spaldin functional if `true`, and Marzari-Vanderbilt for `false`.

- Changed some arrays to allocatable in `disentangle.f90` and `kmesh.f90` for the many k-points computation. (If not, a stack overflow occurs.) (Since this is for personal local tests, it is unnecessary.)

Since the functional sums over k-points before b-vector summation, I needed to order the b-vector information before Wannierization. This sorting is implemented in the `kmesh_sort` subroutine in `kmesh.f90`, and only used for wannier90.x (not postw90.x).

The main changes are in `wannierise.f90` file. `wann_phases`, `wann_omega`, and `wann_domega` subroutines are changed. `guiding_centres=T` case is also considered.

Documentations, tests, and tutorials are added by slightly modifying the silicon tutorial. Also, test data are modified for restart and bvec tests.

Since this is my first pull request in Wannier90, there might be many errors and mistakes. Please feel free to comment.

[1] https://journals.aps.org/prb/abstract/10.1103/PhysRevB.73.075121

PR #504, opened 2024-05-22
Higher-order finite difference

manxkim:
The formula for calculating derivatives in k-space has so far been calculated using the first-order finite-difference(FD) approximation. This PR adds the higher-order FD to make `.nnkp` files. The main changes are in `kmesh.f90.` 

Suppose the first-order FD uses b-vectors. The main strategy of n-th order(n>1) FD is to add 2b, 3b, ... nb vectors, and apply more conditions like Eq. (B1) in [1] to calculate the finite difference coefficients w_b, w_2b, w_3b, ... w_nb.

The scheme is in the same way as we did in one-dimension(see [https://en.wikipedia.org/wiki/Finite_difference_coefficient](https://en.wikipedia.org/wiki/Finite_difference_coefficient))

This PR includes

New parameters:
1. `higher_order_n`: default is 1(1st-order).
2. `higher_order_nearest_shells`: default is `.false.`. If true, the routine uses the nearest shells to satisfy the conditions like (B1). Not extensively tested, and it is an experimental feature.

New tests:
1. `testw90_nnkpt6`: a test for the higher-order bvectors and weights. output: `.nnkp`
2. `testw90_nnkpt7`: a test for the higher-order bvectors and bweights, using the nearest-shells method. output: `.nnkp`
3. `testw90_knbo3_higher`: a test for the Wannier position matrix elements and spreads using higher-order finite difference. output: `.wout`
4. `testpostw90_pt_shc_higher`: a test for the spin hall conductivity using higher-order finite difference. output: `-shc-fermiscan.dat`

[1] N. Marzari and D. Vanderbilt, [Physical Review B 56, 12847 (1997)](https://doi.org/10.1103/PhysRevB.56.12847)


manxkim:
Except the test `testw90_example02_restart`, all tests passed.

The order of b-vectors file has been changed by the subroutine `kmesh_sort`. Therefore `.chk`, and the other outputs (`.mmn`, `.uHu`) including the info of b-vectors should be constructed again. The same issue should be applied to https://github.com/wannier-developers/wannier90/pull/498#issue-2255383555

manxkim:
bvecs in benchmark, and the lines in `.chk` and `.mmn` for lead (testw90_bvec, testw90_example02_restart) have been rearranged.

JeromeCCP9 (approved):
Higher order functionality coded neatly and consistently with new code style.  Tests all pass correctly and new functionality is tested and documented appropriately.

PR #518, opened 2024-09-26
Small fixes for variable allocation

npaulish:
* Some allocatable string variables are not allocated (progname, seedname and cpstatus in wannier_prog.F90, but there might be more for instance in postw90). With stricter compiler settings (e.g. recommended for debugging in [config/make_gfort.inc](https://github.com/wannier-developers/wannier90/blob/dd450c62a0cfd64c34739605a93bfbcdfb0411e0/config/make.inc.gfort#L14)) this causes an error. The problem is caused by `-fno-realloc-lhs` flag. If the arrays are explicitly allocated, the code would be more robust.
*chkpt1 variable used in writing .chk file is declared with undefined length ([wannier90_readwrite.F90](https://github.com/wannier-developers/wannier90/blob/dd450c62a0cfd64c34739605a93bfbcdfb0411e0/src/wannier90_readwrite.F90#L2318)). Parent variable is cpstatus  declared with undefined length in [wannier_prog.F90](https://github.com/wannier-developers/wannier90/blob/dd450c62a0cfd64c34739605a93bfbcdfb0411e0/src/wannier_prog.F90#L74). This breaks backward compatibility with .chk file created with the [old version](https://github.com/wannier-developers/wannier90/blob/1d6b187374a2d50b509e5e79e2cab01a79ff7ce1/src/parameters.F90#L3778) of the code (where len=20). Maybe to ensure backward compatibility it would be better to still use len=20?


JeromeCCP9:
Hello Nataliya!  I found that your change to readwrite.F90 (setting the length to 20) actually breaks reading old checkpoint files, while the allocate-on-assignment style worked fine.  Your change was in the read routine, right?  Could it be that we just want to to make a change to the writing routine and only print (up to) 20 chars there instead?

npaulish:
Thanks Jerome! Sorry for the late reply. Actually once I removed the `-fno-realloc-lhs` flag from the compiler, as we discussed, I could not reproduce the issue with the checkpoint file anymore. Maybe then there is no need to change the routines?

JeromeCCP9:
Hi Nataliya!  Let's indeed close this as unnecessary now.
J.

PR #531, opened 2025-01-15
Large number of updates in preparation for Wannier v4.0

JeromeCCP9:
- large number of minor corrections and fixes
- change from c++ to c header file for library
- changes to (gnu)make system
- first steps in soname-ing of dynamic libs
- moves all writing routines to plot(), eg, write_r2mn, etc
- library document (first draft?)
- projector handling in library case changed
- changes to f90wrap python interface
- various extra tests

In overview, this batch of commits does not complete v4.0, but updates devel with a large number of corrections, cleanups and extensions.


jryates (approved):
Big PR - but I have looked through all of this with Jerome.

PR #548, opened 2025-03-16
Fixes to explicit_kpath code.

jryates:
Fixes a crash on osx+gfortran - the logical has_explicit_kpath was not initialised. This caused the test suite to fail on my MacBook. However, the issue wasn't picked up by the CI.

Also added a check incase explicit_kpath and kpath were both present in the input file.


JeromeCCP9 (approved):
Simple bugfix for uninitialised variable and confused program flow.

PR #549, opened 2025-03-16
Fixes the parsing of f orbital projections

### LINKED ISSUE #432: Error 'param_get_projections: spin is defined but spinors=.false.' for f-orbitals
Dear Developers,
         I am using wannier90-3.1.0 and I found the software complaining 'param_get_projections: spin is defined but spinors=.false.' when I prepared the input files for a spin-polarized DFT calculation involving projection onto f-orbitals. 
        This issue seems to arise becasue the subroutine param_get_projection detects spin direction for projection by scaning brackets ‘(’ and ')' , but there are brackets in the conventional name for f-orbital, i.e. 'fz(x2-y2)', 'fx(x2-3y2)', 'fy(3x2-y2)'. In my case, the program found '(' in f-orbitals and judged that I was using the functionality of spin projection for spinor calculation and report the mentioned error message.  
        If this is the case, the software will also complain 'param_get_projections: found brackets but neither u or d' for spinor calculation with projection onto f-orbitals.

jryates:
This fixes #432 

Three tests have been added.


JeromeCCP9 (approved):
Corrects an edge case tripping up the projector string parser.

Issue #429, opened 2022-07-29
Contradictory statements about CODATA version


jaemolihm:
There are two versions of fundamental constants implemented: CODATA 2010 and 2006.

The header of `constants.F90` says that the default is CODATA 2010,
https://github.com/wannier-developers/wannier90/blob/897a527d473b1a41db094c4a924db700a549a538/src/constants.F90#L26-L29

while `README.install` says CODATA 2006 is the default.
https://github.com/wannier-developers/wannier90/blob/bf0642b53467d63890987831707dbc3e648ef85d/README.install#L223-L229

From the code, it seems that 2006 is the default.
https://github.com/wannier-developers/wannier90/blob/897a527d473b1a41db094c4a924db700a549a538/src/constants.F90#L78-L83

PR #551, opened 2025-03-16
Add support for CODATA 2018 and 2022

jryates:
This addresses #429 and adds support for CODATA 2018 and 2022.

I have left 2006 as the default. If compiled with 2010 or later about 6 of the tests fail. One reason is the conversation from Bohr to Angstrom. All the changes are very small - but larger than our tolerances. We should decide if we want to switch to 2022 and either rebuilt the failing tests - or adjust the tolerances.


JeromeCCP9 (approved):
Corrects a comment about the factually used CODATA version and provides for new versions to be selected.

The default data version is not changed because this changes test data (see issue #562 )

PR #579, opened 2025-11-10
memory bug fix, when more projection than wannier functions are specified

mathunje:
I noticed, that wannier90.x may crash when more projections than Wannier functions are defined and only a subset is selected. I traced back the memory leak via valgrind to the changed line, which initially assumed, that num_wann >= num_proj.


jryates:
Many thanks for reporting this. I can confirm that this is a bug. We've just fixed this in PR #582 . We decided on a slightly different fix - and have also added a suitable regression test. Hence, the new PR and will decline this one.

PR #584, opened 2026-02-02
B sorting

JeromeCCP9:
This PR address the ordering of b-vectors needed for the Stengel-Spalding (wannier90) and translationally-invariant (postw90), removing the need to have differently sorted b-vector lists in these cases.

Previous implementation:  b-vector list different for SS and transl_inv cases, causing m-matrix, uHu, and checkpoint files to be specific to these modes.  Input files generated by the code otherwise are incompatible.

Current implementation: the "old" b-vector sorting method is used always, whether or not SS or transl_inv are used and the input files are unchanged.

b-vector permutation arrays are added to the kmesh_info object.

A problem with guiding centres with the Stengel-Spalding case was revealed: benchmark to be corrected in a separate PR.  This combination is prohibited (if guiding_centers and Stengel Spalding then error exit) until fixed.


JeromeCCP9:
This PR concerns the issues discussed in #539 where a b-vector sorting option was suggested instead.

JeromeCCP9:
@sjhong6230 @manxkim 

Dear Seung-Yu and Minsu,  here I push the workaround b-vector sorting that avoids having to change the input files.  Of course this means that we have to use a permutation array whenever we do the b-vector sums (in use_ss_functional or transl_inv_full cases).

I think the big advantage of this is that it is easier for the user because they can simply switch these methods on/off without regenerating inputs.

The permutation is pretty simple in the code; I'm not sure how optimal it is.

I temporarily commented the higher_order SS test, actually because it is SS+guiding centres: there is an issue with the existing benchmark that we can discuss in a separate PR soon.

Thank you both so much for your help with this!!
Yours, Jerome

jryates (approved):

PR #589, opened 2026-02-04
make allocation of large arrays czmat_in/out conditional 

jryates:
make allocation of large arrays czmat_in/out conditional on lsitesymmetry - as they are now only used with the site symmetry path.


JeromeCCP9 (approved):
Simple, clear improvement.

PR #592, opened 2026-02-09
avoid Intel/GCC BLAS ABI incompatibility by avoiding blas functions

JeromeCCP9:
i.e. incorrect behaviour when ifx combined with (GCC compiled) Openblas

fix replaces zdotc function (return result passing being the problem) with zgemv subroutine


jryates (approved):

LecrisUT:
Cool find and fix 👍

PR #595, opened 2026-02-16
Performance optimizations for plotting cube files

elinscott:
This PR contains a few minor changes that greatly introduce the speed of plotting cube files and a few related bugfixes.

### Optimizations
+ **Band contraction** Precompute `c_wvfn(npoint, w) = sum_b u_matrix * r_wvfn` once per unit cell grid per k-point. The supercell loop reuses this instead of recomputing the band sum at each replica.
+ **Phase factoring** Precompute phase per k-point, replacing per-grid-point `exp()`
+ **Spinor simplification** The k-loop now only accumulates into `wann_func_nc`; the redundant per-k-point spinor to `wann_func` conversion is removed.
+ **Batched cube write** with 6 values per write statement instead of one-at-a-time with `advance='no'`.

### Bug fixes
+ Table lookup of atomic Z was failing due to the fact `atom_data%symbol` has been changed to be lowercase; the lookup table now is also lowercase to match
+ XSF and XYZ files now use title-cased labels instead of lowercase symbols.

### Tidying
+ Shorthand variables introduced for the bounds of the supercell grid indices


elinscott:

JeromeCCP9 (approved):
Nice, clean enhancements to the code.

PR #599, opened 2026-03-16
read seedname.kshell file when input token kmesh_shell_from_file = T

JeromeCCP9:
Replace old "devel" flag & select b-vectors by hand as fix in specific cases using the ".kshell" file.

Fixes issues 436 and 437.

This restores the functionality that "devel_flag=kmesh_degen" provided in version 3.1.

Issue 436 is added as a test case, but it is necessary also to specify "skip_b1_tests" in order to execute successfully; since this was also required for version 3.1.0, I guess this is correct behaviour?


jryates:
I don't recall needing to set skip_b1_tests in 3.1 - indeed that keyword was added quite recently, and the devel_code fix predates this. I will check this.

JeromeCCP9:
Hello Jonathan--I looked into this again and found that there were two implementations testing the B1 condition.  One is the usual code that is executed routinely in the main program flow (kmesh.F90 around line 594) and an *additional* check specific to the read of the .kshells file, in subroutine kmesh_shell_from_file().  Both tests were protected by the skip_B1_tests flag.

Because the test in kmesh_shell_from_file() was encountered first, this code was causing the need for setting "skip_B1_tests=T" workaround.

I removed this initial test entirely, allowing the code to proceed to the usual B1 condition test, which passes without any issue.  Therefore I removed the "skip_B1_tests=T" workaround from the test also.

I didn't debug the difference between the "usual" B1 test and the one in kmesh_shell_from_file()--the latter is much older code and possibly just buggy.  I guess it's not worth debugging this if the "usual" B1 test is robust.  Since it happens essentially at the end of the kmesh setup procedure, I guess we're good if it's ok.

JeromeCCP9:
I checked that the same applies to the BN problem also, but I didn't add this as a test case since it really duplicates the GaP one.

jryates:
It's not obvious to me why kmesh_shell_from_file checks the B1 condition - given it is done later on. It is fine to skip it.

JeromeCCP9:
yes, I thought similarly.  I'm happy that this resolves the issues now.

jryates (approved):

PR #431, opened 2022-09-30
Tetrahedron method to calculate spin Hall conductivity

manxkim:
I am going to add the tetrahedron method to postw90.
First, I implemented the tetrahedron method to calculate spin Hall conductivity using the tetrahedron method suggested in my work, [PRB 106, 075126](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.106.075126). The tetrahedron method will eventually be extended to other quantities, such as ahc. (or other k-integration of spectral functional forms)

Three parameters, two tests, and one module `Tetrahedron.F90` have been added, and `Berry.F90` has been mainly modified. One thing is that there are some parts regarding tetrahedron in `berry.F90`, so it also should be moved to `Tetrahedron.F90`.

PR #600, opened 2026-03-19
Tetrahedron

JeromeCCP9:
Replacement PR after slightly complex rebase.

Original comment: 

[manxkim](https://github.com/manxkim)
commented
[on Sep 30, 2022](https://github.com/wannier-developers/wannier90/pull/431#issue-1392133709)

I am going to add the tetrahedron method to postw90.
First, I implemented the tetrahedron method to calculate spin Hall conductivity using the tetrahedron method suggested in my work, [PRB 106, 075126](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.106.075126). The tetrahedron method will eventually be extended to other quantities, such as ahc. (or other k-integration of spectral functional forms)

Three parameters, two tests, and one module Tetrahedron.F90 have been added, and Berry.F90 has been mainly modified. One thing is that there are some parts regarding tetrahedron in berry.F90, so it also should be moved to Tetrahedron.F90.


JeromeCCP9:
Dear @manxkim !

I added the tetrahedron method.  But I found some issue with the test sets.  Please can you check them?

- commit #431 0da964b4 doesn't reproduce the benchmark data contained in that commit (on my vanilla gcc setup here), can you check this, please?

- the rebased version (here) also doesn't reproduce the benchmark data.

The differences are small, but enough to fail the test.  Seemingly the real part of SHC is most affected.  If your tests pass for you on commit 0da964b4, then I'll go back and try to find where the discrepancy is coming from.

Thank you for your very nice code here.
Yours, Jerome

JeromeCCP9:
One thing that stands out in the test:

testpostw90_pt_tetra_shcfermi/

is that the function in shc-fermiscan.dat has an extremely large value at one frequency point, orders of magnitude larger than the variation across the other frequencies.  When you use sampling instead of tet, this spike is not present.

J.

manxkim:
Dear @JeromeCCP9 , sorry for the late comment.

I found where the discrepancy comes mainly from, and it indeed is highly likely to be a problem regarding the numerical (precision) error mentioned in my paper, [https://doi.org/10.1103/PhysRevB.106.075126].
I tested two systems: a desktop and my laptop. They produced different data and didn't pass the test: for example, (testpostw90_pt_tetra_shcfreq)

1)
ERROR: absolute error 8.31e-01 greater than 1.00e-01. (Test: 1576.4265.  Benchmark: 1577.2574.)
ERROR: absolute error 1.65e-01 greater than 1.00e-01. (Test: 1648.1135.  Benchmark: 1648.2785.)
ERROR: absolute error 1.23e-01 greater than 1.00e-01. (Test: 2291.9208.  Benchmark: 2291.798.)

2)
ERROR: absolute error 1.86e-01 greater than 1.00e-01. (Test: 1577.443.  Benchmark: 1577.2574.)
ERROR: absolute error 3.20e-01 greater than 1.00e-01. (Test: 1592.1388.  Benchmark: 1592.4584.)
ERROR: absolute error 2.64e-01 greater than 1.00e-01. (Test: 1648.014.  Benchmark: 1648.2785.)

(Compare 1576.4265 and 1577.443)


As you mentioned, "the differences are small, but enough to fail the test."
This error is a natural artifact of the form of the integrals: if one calculates log(1+x)/x (even with the more precise log1p), sufficient care must be taken. Otherwise, random numerical error can be pronounced. To address this problem, the `tetrahedron_cutoff` parameter is set.

I originally set it to 1.e-4, but it was not enough to produce consistent results across several environments.
Even though I suggested 1.e-3 in my work, I mistakenly wrote 1.e-4 in the win file.
As I changed it to be a greater 1.e-1, the discrepancy disappeared:

1)
ERROR: absolute error 1.85e-01 greater than 1.00e-01. (Test: 1461.2048.  Benchmark: 1461.3901.)
ERROR: absolute error 1.81e+00 greater than 1.00e-01. (Test: 1575.4523.  Benchmark: 1577.2574.)
ERROR: absolute error 1.35e+00 greater than 1.00e-01. (Test: 1591.1086.  Benchmark: 1592.4584.)

2)
ERROR: absolute error 1.85e-01 greater than 1.00e-01. (Test: 1461.2048.  Benchmark: 1461.3901.)
ERROR: absolute error 1.81e+00 greater than 1.00e-01. (Test: 1575.4522.  Benchmark: 1577.2574.)
ERROR: absolute error 1.35e+00 greater than 1.00e-01. (Test: 1591.1086.  Benchmark: 1592.4584.)

I'm going to look further into the error dependence on tetrahedron_cutoff in this case and increase the cutoff or tolerance. I'll make a commit.

Thank you for your efforts on this issue.

Best, Minsu

JeromeCCP9:
Dear Minsu, this is very nice news!

I see that a modification to the test tolerance will likely solve the problem--then we can certainly merge this shortly!  This is great!

One thing--the test case has a conspicuous spike at one frequency (exactly) that is not present in sampling--what do you think about this?  I wondered whether it was an unfortunate edge case: perhaps a degeneracy of the evaluation frequency with one or more of the tetrahedon energies?

Thank you for looking at this so quickly after making the original PR so long ago.
Yours, Jerome

manxkim:
Dear Jerome,

As I checked, it is near E = 11.8999-11.9000 eV, right?
I found a peak here, as below,

#No.   Fermi energy(eV)   SHC((hbar/e)*S/cm)
...
  10    11.899890   -0.39680245E+03
  11    11.899900   -0.39702473E+03
  12    11.899910    0.27640585E+05
  13    11.899920    0.27653485E+05
  14    11.899930    0.27666388E+05
  15    11.899940    0.27679296E+05
  16    11.899950    0.27692208E+05
  17    11.899960    0.27705124E+05
  18    11.899970    0.27718045E+05
  19    11.899980    0.27730968E+05
  20    11.899990    0.27743895E+05
  21    11.900000    0.27757034E+05
  22    11.900010   -0.18890245E+03
  23    11.900020   -0.18890341E+03
  24    11.900030   -0.18890768E+03
...

As you mentioned, this has to do with a degeneracy. This is because there's a set of not-well-treated degenerate bands. In lines 302-303 of `tetrahedron.f90`, I originally set a small parameter to avoid this problem,
```
    !small parameter to avoid a problem of degenearcy
    real(kind=dp), parameter :: avoid_deg = 1.e-4_dp
```
, which exactly corresponds to the range (11.8999, 11.9000]. This value, 1.e-4, was not sufficient.
The culprit bands are not originally near 11.9000 eV but are shifted here due to the tetrahedron_higher_correction = true option, which modifies the energy eigenvalues (and also the spin-current matrix elements). When I turned off this option, the spike disappeared.

I checked that 3.e-4_dp was enough to remove the spike in our case, but generally speaking, it depends on a material's band structure. If one does fermiscan, avoid_deg needs to be set properly. Therefore, I think it's better to switch this hard-coded parameter `avoid_deg` into an external one with a default of 3.e-4_dp, for example, 'tetrahedron_avoid_degeneracy'. (degeneracy would be better than deg, which can be confused with degree) What do you think about this? If it seems to be okay, I'll add a new parameter.

Thanks!

Best, Minsu

manxkim:
p.s. These frequencies below seem too many, so I am going to reduce them to (13, 23, 0.5), which is about the Fermi energy plus  minus 5.
```
fermi_energy_min = 6
fermi_energy_max = 26
fermi_energy_step = 0.1
```

JeromeCCP9:
Dear Minsu,  this is really great!  Thank you for attending to this & adding the new flag.  

(edit)  I rebased on develop here and I think this is now ready to merge.
Thank you again,
J.

JeromeCCP9:
@manxkim the tests seem to show some numerical discrepency when using intel's MKL library:

shc_re
    ERROR: absolute error 1.38e-01 greater than 1.00e-01. (Test: 1646.7136.  Benchmark: 1646.8514.)

Do you think this is within acceptable range, and that we should just lower the tolerance, or does this need investigation?

Sorry about this delay!
Yours, Jerome

manxkim:
I thought it might be the case, and it looks like 1.00e-01 was a bit tight.
It's okay to increase the tolerance to 1.00e-00.

Thanks!

Best,
Minsu

jryates (approved):

PR #605, opened 2026-03-19
test strange yaml err

JeromeCCP9:
(debugging a yaml/CI config error)

@LecrisUT  Hello again, Cristian!

strangely, inputs.mask-experimental doesn't seem to be defined in step_test.yaml -- could you take a quick look, please?  Unfortunately it's jammed the CI.

### DIFF
--- a/.github/workflows/step_test.yaml
+++ b/.github/workflows/step_test.yaml
@@ -26,6 +26,7 @@ jobs:
       matrix:
         toolchain: [ gcc ]
         mpi: [false, openmpi, mpich]
+        experimental: [false]
         include:
           - toolchain: intel
             mpi: false


LecrisUT:
That is news to me, it should just expand to nothing/empty string if not defined. Any link for where it failed?

LecrisUT:
And it is defined and passed so I'm doubly confused how something might have broken

https://github.com/wannier-developers/wannier90/blob/23f6cc47a6fa8d2ede4a1a86374dfa22ec6a9bcd/.github/workflows/ci.yaml#L30

JeromeCCP9:
here is the problem in action: https://github.com/wannier-developers/wannier90/actions/runs/23314121761/job/67810544761?pr=595

does ${{''}} not evaluate to a bool?

LecrisUT:
Ah, I think I understand and the issue is probably the other way arround. Can you try to set `experimental: [false]` under `matrix` key over there?

JeromeCCP9:
nice indeed--well figured out.

JeromeCCP9:
do I remember correctly that we marked the intel tests as "experimental" because they were failing?  Now that they're fixed, there isn't really a reason to mark them "experimental" (apart from the general badness of linking fortran compiled by different compilers!)--should we remove that attribute from the intel ones?

jryates (approved):

LecrisUT:
> should we remove that attribute from the intel ones?

Sounds good :). Reminds me that I need to update that image soon to get the latest version.

Hmm, did we not get to enabling the packit app to build on the latest Fedora?

PR #610, opened 2026-03-24
add .win,.mmn,.amn printout option "dump_files"

JeromeCCP9:
provides the ability to dump input file (.win) and overlap, projection, eigenvalue files from data entered via the library interface.

triggered by setting the option "dump_inputs=T" which is only meaningful in the library case.


jryates:
ierr is not set in w90_print_win

jryates (approved):

PR #612, opened 2026-03-25
cmake: add MPIH build option for legacy mpi.h interface

the-hampel:
Allow CMake builds to select the legacy mpif.h MPI interface instead of the default MPI90 module path, and apply the GNU Fortran compatibility flags needed for that configuration to build successfully.


JeromeCCP9 (approved):
adds mpi.h support and some gcc arg mismatch flags.

the-hampel:

JeromeCCP9:
recheck

PR #614, opened 2026-03-25
Rewriting the (old) Makefile system, fix `make install`

giovannipizzi:
The old one was very complicated, as it was originally designed when postw90.x was introduced, but only postw90.x was parallel and wannier90.x was not.
So, it has a very convoluted way for different makefiles to depend on each other, making the definition of variables in a single place very cumbersome.

This rewrites the whole system making it more straightforward.

Moreover, it removes old relics (e.g. the old library), renames make targets more sensibly (staticlib and dynlib, and libs for both), and fixes the `make install` to fix (hopefully correctly!) the .pc file for pkgconfig, and install also the libraries and the .mod files.


JeromeCCP9 (approved):
Cleanup of make system with lots of clarifications, various fixes, etc.

PR #617, opened 2026-03-26
cmake: fix library when using cmake with MPI

the-hampel:
the normal build system builds libwannier90_mpi.a/so when using MPI and libwannier90.a/so for the serial build. This was not the case for the cmake build. This commit fixes this.

### DIFF
--- a/.github/workflows/step_test.yaml
+++ b/.github/workflows/step_test.yaml
@@ -51,6 +51,9 @@ jobs:
 
           diff orig_env module_env | sed -n 's/> //p' >> $GITHUB_ENV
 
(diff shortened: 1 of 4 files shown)


JeromeCCP9 (approved):
fixes known bug in cmake config

JeromeCCP9:
This was a small can of worms--the github cmake workflows were not previously doing parallel tests at all, and activating them in this commit revealed some other problems.

Once rebased on #632 , this should test successfully and be merge-able.
J.

the-hampel:
Oh wow I did not realize at all. Very good find! Fingers crossed.

PR #624, opened 2026-03-27
Provide exclude bands list

JeromeCCP9:
adds another library function to provide the list of excluded bands after parsing the W90 input/options.


JeromeCCP9:
@jryates this can be merged; I'll update the documentation with this new facility in a following PR (doc updates together).

jryates (approved):

PR #629, opened 2026-03-31
Improve C interface

qiaojunfeng:
- Cleaned up C API functions, e.g., projections and exclude bands
- Moved wrapper implementations out of the header into a dedicated C source file, so that wrapper functions are properly compiled in the lib.so/.a file
- Updated the C++ example
- Few bug fixes in win parser

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -44,6 +44,7 @@ target_sources(Wannier90_lib PRIVATE
 		transport.F90
 		types.F90
 		utility.F90
+		wannier90.c
 		wannier90_readwrite.F90
 		wannier90_types.F90
 		wannierise.F90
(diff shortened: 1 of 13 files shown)


qiaojunfeng:
This also contains a fix for #628

JeromeCCP9 (approved):
This is a lot of work, Junfeng; very welcome improvements and fixes.

If this fixes the projector reading, that is also very nice.  Let's also add the test from Henrique for that.

qiaojunfeng:
Hi @henriquemiranda , I implemented a bit different fix in this PR, and refactored your bug report in #628 as a ctest in
`test-suite/library-mode-test-projections/set_projections.f90`. I also added you as a co-author in the git commit, if this is OK with you, maybe we can close #628? Otherwise, let me know and I can move my last commit to your PR. Thanks!

henriquemiranda:
Sure no problem! Just tell me, did you reproduce the issue?

qiaojunfeng:
> Sure no problem! Just tell me, did you reproduce the issue?

Yes I was facing the same bug last week while testing the library mode.

henriquemiranda:
I would swear that it worked at some point but I don't know anymore. Anyway, thanks for having a look and fixing it! I will close #628

JeromeCCP9:
This could have been exposed by commit e1f65ec that I added last week; prior to that, the counting of entries was incorrect by one in a way that could have hidden this problem.  It's very good to have a test!  Thank you both for fixing this!  J.

PR #632, opened 2026-04-08
Workflows mpi fixups

JeromeCCP9:
this provides some fixes in advance of #617  (#617 should be rebased on this).

#617  reveals that the cmake parallel test workflow was incorrect by 1. not setting WANNIER90_MPI=ON and 2. by not correctly setting up the PATH after loading mpi modules.

Correcting the parallel switch (in 617) and correcting the PATH (in this commit) reveals that the GCC+MPICH test is misconfigured: it hangs forever on the first mpirun.  I believe this is an MPICH specific problem in the workflow definition (because intel and openmpi MPI work correctly); therefore I disabled the MPICH tests.

### DIFF
--- a/.github/workflows/step_test.yaml
+++ b/.github/workflows/step_test.yaml
@@ -51,6 +51,9 @@ jobs:
 
           diff orig_env module_env | sed -n 's/> //p' >> $GITHUB_ENV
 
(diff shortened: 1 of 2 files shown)


JeromeCCP9:
closed because commits added via 617 instead.

PR #640, opened 2026-04-30
Propagate staticlib target from top level Makefile

DBJochym:
This fixes the case of a requested `staticlib` GNU Make target not being propagated. Without the specified target, the behaviour is to default to the first target in `src/obj/Makefile`, which at time of writing is `wannier`.


DBJochym:

JeromeCCP9 (approved):

PR #645, opened 2026-06-15
Remove polyfill-dot-io from docs

### LINKED ISSUE #643: polyfill login prompt pops up when accessing documentation
Image

When I access https://wannier90.readthedocs.io/en/latest/, I see the above popup. (The text left to the boxes are in Korean and means "Username" and "Password")

I have no knowledge of web browsers or JavaScript, but it seems the following line is relevant.
https://github.com/wannier-developers/wannier90/blob/1bbde1332bcc4147751e499a1c294cb1a02a9e04/docs/mkdocs.yml#L224

jryates:
This was needed to allow IE11 users to see formatted equations. It is now a security risk. Fixes #643

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -221,7 +221,6 @@ plugins:
 
 extra_javascript:
   - javascripts/mathjax.js
-  - https://polyfill.io/v3/polyfill.min.js?features=es6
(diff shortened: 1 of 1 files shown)


qiaojunfeng:
Thanks! Yes it is mostly for back compat which is probably not needed nowadays. Removing that and the docs render fine on both chrome and firefox, plus the benefit of removing a security hole.

qiaojunfeng (approved):

PR #647, opened 2026-07-06
Fix MPI hang/abort on UNK read failure in `plot_wannier`

elinscott:
## Symptom
In `wannier_plot`, a missing, truncated, or corrupted UNK file kills the run with a raw Fortran runtime error and no wannier90 error message. A missing UNK file is silently created empty in the process.

This can be reproduced using the content of `testw90_cube_format`: delete `UNK00005.1` and run. The code writes a 0-byte `UNK00005.1`, then dies with `Fortran runtime error: End of file` and a backtrace.

## Cause
The UNK opens in `plot_wannier` had no `status='old'`, so opening a missing file created it. None of the header/wavefunction reads had `iostat` guards, so a malformed file caused an ugly abort.

### Earlier misdiagnosis
Earlier commits and review discussion wrongly attributed this failure to mismatched MPI collectives. However, the comms/error layer already handles a lone failing rank by design. The final commit reverts the error plumbing to the standard "return" idiom.

## Changes
- All UNK opens now have `status='old'` plus an `inquire` existence check, so a missing file produces a proper error
- All UNK reads carry `iostat` and report failures through the standard `set_error_file` mechanism with an immediate return upon an error
- The per-k-point read sequence is factored into a helper
- the previously duplicated disentangled/direct and spinor read paths are unified
- The `UNK00001` grid-dimension probe at the top of `plot_wannier` gets the same `status='old'` & `iostat` checks (a truncated `UNK00001` also previously produced a raw abort)

### DIFF
--- a/src/plot.F90
+++ b/src/plot.F90
@@ -1558,9 +1558,9 @@ contains
     use w90_types, only: wannier_data_type, atom_data_type, dis_manifold_type, print_output_type, &
                          timer_list_type
     use w90_wannier90_types, only: wvfn_read_type, wannier_plot_type
-    use w90_comms, only: w90_comm_type, comms_sync_error
+    use w90_comms, only: w90_comm_type
     use w90_error, only: w90_error_type, set_error_alloc, set_error_dealloc, set_error_file, &
-                         set_error_warn, set_base_error, code_file
+                         set_error_warn
 
     implicit none
 
@@ -1613,8 +1613,6 @@ contains
     integer :: loop_kpt, nk, ngx, ngy, ngz, nxx, nyy, nzz
     integer :: loop_b, nx, ny, nz, npoint, file_unit, loop_w, num_inc
     integer :: wann_plot_num
-
-    integer :: sync_code
 
     character(len=11) :: wfnname
     character(len=60) :: wanxsf, wancube
@@ -1755,7 +1753,7 @@ contains
       end if
 
       call io_date(cdate, ctime)
-      kpt_loop: do loop_kpt = 1, num_kpts
+      do loop_kpt = 1, num_kpts
         if (dist_k(loop_kpt) /= my_node_id) cycle
 
         inc_band = .true.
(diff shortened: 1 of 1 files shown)


JeromeCCP9:
This is nice!  May I modify it slightly to put the set_error_* calls closer to the error condition along with immediate returns (instead of waiting until the end of the subroutine); that makes the code a little simpler and matches the way we check errors elsewhere (eg in deallocs, etc)?

elinscott:
> This is nice! May I modify it slightly to put the set_error_* calls closer to the error condition along with immediate returns (instead of waiting until the end of the subroutine); that makes the code a little simpler and matches the way we check errors elsewhere (eg in deallocs, etc)?

I've tweaked, it, as far as MPI allows, but the return itself can't be moved. Take `mpirun -np 2` with 8 k-points, where rank 1 owns `UNK00005.1`, and that file is missing. Currently, `set_error_file` contains an `mpi_allreduce` (via `comms_sync_error`), so if rank 1 called it and returned at the failure site, rank 1 would sit in that `allreduce` while rank 0 - which saw nothing wrong - proceeds to the `comms_reduce` of `wann_func` and hangs. To avoid this, we need to wait until all ranks meet at a single `comms_sync_error`.

Cf. the error checks for deallocate you mentioned: these can do an immediate return because all ranks run the same `deallocate`, so in practice* either all or no ranks enter the sync inside `set_error_dealloc`.

I was able to tidy things a little using `set_base_error`, but I now realise that this would be the first use of this funciton outside of `comms.F90` — possibly something you might want to avoid?

*a `deallocate` really only fails on programming errors such as deallocating an array that was never allocated, and so all ranks hit the same bug

elinscott:
... actually, thinking about it a bit more this what we need is a helper `read_unk` function within the kpt loop. Then things become a lot neater. Let me implement that...

elinscott:
... done!

JeromeCCP9:
I think it would be ok to 'return' from within the k-point loop, no?  (It's just a small question, I think we can merge this already.)

elinscott:
No — if failing ranks return from within the k-point loop and succeeding ones do not, a combination of failing and succeeding ranks would encounter different comms routines (which was the reason for the original ugly crash).

That said, the refactor into a helper routine achieves the effect that you're after with an in-loop return: inside the helper each failure returns immediately, rather than having ierr checks everywhere.

JeromeCCP9:
We designed the comms/error such that this would exactly not be a problem--the next collective operation (regardless of what it is!) is when the error status is propagated.  If what you write is true, that means there is a larger problem; may I ask, did you test it, actually?  May I test it?
J.

elinscott:
@JeromeCCP9 ah, you're right, sorry! Having another look at it, the real issue was the lack of a `status=old` and `iostat` checks throughout the UNK reading. These were added at the same time as me trying to catch all the branches with the same `comms_sync_error`, so when the problem vanished during my testing I misattributed it to me routing every rank through the same comms call.

I have updated the PR description — have a read for more details (including how to reproduce).

I have also updated the PR content so that it uses the immediate return as you suggested and the error handling remains clean.

Thanks for the catch!

JeromeCCP9 (approved):
Adds error checking/handling (using the current design) for reading of UNK files.

Slightly verbose because the file read is encapsulated in a new function, which counts as an improvement.

Comments are in the "claude fable5 style", but all otherwise seems fine.

(Unfortunately, this functionality is not explicitly tested in the test suite right now.)

jryates (approved):

PR #649, opened 2026-07-12
collection of small fixes for v4

JeromeCCP9:
Some tasks are still outstanding:

[] issue 634 needs better documentation
[] CHANGELOG should be reviewed
[] issue 464 (CONTRIBUTING.md should be updated)

only 5 tests were caused to fail due to the CODATA version change, so I replaced only those benchmarks.

serial tests now pass for my nvfortran (nvfortran 26.5-0 64-bit target on x86-64 Linux -tp haswell)


jryates (approved):

JeromeCCP9:
(634, 464 and documentation updates to follow in a separate PR.)

PR #653, opened 2026-07-19
fix(cmake): gate C interface behind WANNIER90_WITH_C

jimustafa:
`README.install` mentions using `-DWANNIER90_WITH_C=ON` to build the C interface, but the option does not appear to be defined and used to gate inclusion of the C interface sources and install of the C interface.


jimustafa:

JeromeCCP9 (approved):
Thank you so much for this work!

PR #668, opened 2026-08-21
Update CHANGELOG.md with correction to license update

JeromeCCP9:
"or later" was missing in the license update description


mostofi (approved):
Thanks for updating this.

Issue #671, opened 2026-08-27
Library interface - Dump_input : Segfault when wannier90 is called in parallel line 464.


durhamb-york:
When dump_input is set to true running via the library interface, and the library is called in parallel, the library segfaults on line 456 on the 4th dimension of the `m_matrix`.

Examining the code, the routine `overlap_write` clearly assumes it has all of the M matrices on the root node. This is easily avoided by performing a reduction to get the global data all on the root node.

Seg fault message: 
At line 456 of file ../overlap.F90
Fortran runtime error: Index '8' of dimension 4 of array 'm_matrix' above upper bound of 7

Error termination. Backtrace:
#0  0x5add397887c2 in ???
#1  0x5add39788a59 in ???
#2  0x5add39789046 in ???
#3  0x5add394b8784 in __w90_overlap_MOD_overlap_write
	at ../overlap.F90:456
#4  0x5add394a665d in __w90_library_MOD_w90_disentangle
	at ../library_interface.F90:584

JeromeCCP9:
Hello Ben!  Thank you for reporting this, I'll check & reply again.  Can you give me the git hash of the version you're running, meanwhile, please?  J.

durhamb-york:
Hi Jerome, I've just tested and it still fails on the latest HEAD so commit hash 7129b45e8. I have a fix I'm preparing as well, but thought I would raise the issue first.

JeromeCCP9:
Thank you!  If you are preparing a solution, then I guess I shouldn't work on this?  I'll wait for your PR.

durhamb-york:
Hi Jerome, see PR #672 for my proposed fix. Ben

JeromeCCP9:
fixed in #672

PR #672, opened 2026-08-27
Avoid segfault in overlap_write when called in parallel

durhamb-york:
Fixes issue #671 of seg faults when w90 is called in library mode in parallel with the dump_input flag set to true.

When writing the M matrices out to file, we first have to do a reduction to get the data from all the k-points on the root node.


durhamb-york:

JeromeCCP9:
m_matrix is always the coalesced version (not the decomposed one)... let me track down why this isn't necessarily/apparently available & maybe iterate this with you.  Brb.  J.

JeromeCCP9 (approved):

PR #679, opened 2026-09-08
Add routine that sets MPI communicator type via MPI_VAL member

durhamb-york:
Adds a routine to the library interface that allows the MPI communitcator to be set agnostic of how the library and calling program were compiled, with MPI types or not. This means that the calling program can have clean higher level code which handes the Wannier90 interface which can be completely agnostic about how/if it was compiled with MPI, and is not passing around MPI types.

### DIFF
--- a/src/library_interface.F90
+++ b/src/library_interface.F90
@@ -227,6 +227,8 @@ module w90_library
   !! transform overlaps and initial projections
   public :: w90_set_comm
   !! setup MPI communicator in parallel case
+  public :: w90_set_comm_integer
+  !! setup MPI communicator in parallel case via integer handle to MPI comm
   public :: w90_set_constant_bohr_to_ang
   !! set value of Bohr/Angstrom conversion
   public :: w90_set_eigval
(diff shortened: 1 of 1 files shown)


JeromeCCP9:
Hi Ben!

Can you explain this a bit more?  You would like to decouple the style of MPI used by the library and the calling program, I guess?

J.

durhamb-york:
Hi Jerome, the aim is partly to decouple the MPI style of the calling program and library, but also to remove the need to for the MPI comms type to be exposed in high level code in the calling program.

For CASTEP, we would quite like to keep our interface with the MPI contained to one module, and not expose any MPI comms types in the high level code.  At the moment this means we enforce Wannier90 to be built using the older style, which would potentially prevent us using a precompiled library version.

The added routine provides a more flexible interface, so we when compiling the calling program, we don't need to know how wannier90 was compiled or force the compilation to use the less modern interface.  Since the MPI_VAL member is designed for this type of interoprebility (see https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node25.htm ), I thought it made sense to propose this addition. 

Ben

JeromeCCP9:
Thank you for your reply!  This is a kind of workaround for inconsistent compilation, I think--is this really blocking you or is this a "nice to have" kind of thing?  The additional function adds complexity, which is a downside.

durhamb-york:
It's definitely more a nice to have, rather than a blocker.  I will say that it greatly simplifies the interactions between the CASTEP and Wannier build system and removes the possiblity of an compile error where a calling program is compiled with an older style of MPI  to the library.

Issue #654, opened 2026-07-20
sitesym_symmetrize_gradient argument cdq incorrect


JeromeCCP9:
src/wannierise.F90 +600

- cdq passed but only cdq_loc meaningful here
- incorrect dimensioning of cdq_loc in the parallel case
- tests pass in serial and parallel (misleading!?)
- seems to predate v3.1 (tbc)

JeromeCCP9:
Replacing this issue with a more general one about SAWF testing (#728 )

PR #681, opened 2026-09-08
correct latent bug in pasing cdq instead of cdq_loc in sitesym_gradient

JeromeCCP9:
(this has extremely small numerical effect and tests are unaffected)


JeromeCCP9:
fixes #654

src/wannierise.F90
```diff
@@ -599,7 +599,19 @@ subroutine wann_main(ham_logical, kmesh_info, kpt_latt, wann_control, omega, sit
       if (allocated(error)) return
 
       if (lsitesymmetry) then
+        ! symmetrize_graident requires all k (IBZ and related FBZ points for each op)
+        cdq(:, :, :) = 0.0_dp
+        do nkp_loc = 1, nkrank
+          nkp = global_k(nkp_loc)
+          cdq(:, :, nkp) = cdq_loc(:, :, nkp_loc)
+        end do
+        call comms_allreduce(cdq(1, 1, 1), num_wann*num_wann*num_kpts, 'SUM', error, comm)
```
jryates, line 608: This should have an test for the error.

jryates (approved):
This code looks correct. However, I am yet to find a test in which the fix makes a difference. i.e. the symmetrisation of the gradient does not appear to matter. Perhaps this is the way things are. I will open an issue to record this - but the PR should be accepted.

PR #689, opened 2026-09-09
rename MPI preprocessor symbols to W90_MPIx (resolve some mpich issues)

### LINKED ISSUE #665: Fedora packaging: `-DMPI` is not public
I have not run some proper post-packaging tests and it seems I did not add any for this project, maybe will contribute some later. But at first glance, it seems that the lack of the define would be problematic for the header files:
https://github.com/wannier-developers/wannier90/blob/e9f448e1ff4199ab5f983ec7265adca06b6363e8/src/CMakeLists.txt#L90-L93
https://github.com/wannier-developers/wannier90/blob/e9f448e1ff4199ab5f983ec7265adca06b6363e8/cmake/wannier90.pc.in#L11-L12
https://github.com/wannier-developers/wannier90/blob/e9f448e1ff4199ab5f983ec7265adca06b6363e8/src/wannier90.h#L78-L84

At the same time, populating that for other projects would also be problematic, so maybe that macro guard could be changed to something like `-DWANNIER90_MPI` or such.

Fortran mod files should be fine though I guess.

JeromeCCP9:
fixes #665


JeromeCCP9:
@LecrisUT I renamed the symbols to avoid -DMPI; I think this should resolve your mpich issue?

jryates (approved):

LecrisUT:
Hmm, on paper looks right. I'll let you know when I backport it on the Fedora review (seems like I still haven't submitted it)

PR #693, opened 2026-09-09
Replace the testcode test suite with pytest

giovannipizzi:
## Goal

The functional test suite was driven by `testcode2`, a third-party tool vendored into
`test-suite/testcode/`, frozen at a 2017 version and patched by hand. It is now
unmaintained. This PR replaces it with pytest, and takes the opportunity to fix
a few things that made the test suite awkward to work with.

The work is split into individual, task-oriented commits (13 of them), each one
self-contained and separately reviewable.

**No test input, reference or checkpoint file has had its content changed.** Reference files
were renamed, but every one of the 97 is recorded in git as a pure rename with zero changed
lines. The new suite gives exactly the same pass/fail results as the old one.

## What changes for someone running the tests

- `./run_tests` and `./clean_tests` are gone. Run `pytest` from `test-suite/` instead.
- Python 3.10 or newer is now required, plus `pytest` and `PyYAML`
  (`pip install -r test-suite/requirements.txt`).
- `make tests`, `make test-serial`, `make test-parallel` and `ctest` all still work and now
  go through pytest.

Common commands:

| Command | Does |
|---|---|
| `pytest` | run everything, serially |
| `pytest -k testw90_example01` | run one test |
| `pytest -m postw90` | run only the postw90 tests |
| `pytest --nprocs=2` | run under MPI |
| `pytest -n auto` | run in parallel across cores (about 5x faster) |
| `pytest --update-benchmarks` | regenerate reference files |

### Tests now run outside the source tree

Each test is copied to a temporary work directory and run there. The source tree is never
written to, so `git status` stays clean, runs cannot contaminate each other, and there is
nothing left to clean up afterwards (hence no `clean_tests`). `--workdir` and
`--keep-workdir` let you keep the files around to inspect a failure.

### Test configuration is per-test and declarative

The two central config files (`tests/jobconfig` and `tests/userconfig`) are replaced by:

- `tests/<name>/test.yaml` — one small file per test, saying which input to run, what output
  to compare and which profile to compare it with. The descriptive comments that used to sit
  in `jobconfig` are preserved here, next to the test they describe.
- `test-suite/profiles.yaml` — the 30 comparison profiles (which parser to use, which
  tolerances to apply).

The per-test `Makefile`s (55 of them, used to uncompress inputs and convert checkpoints) are
replaced by a `prepare:` list in `test.yaml`. Everything a test needs is now described in one
place.

### Adding a test is a single step

Create the directory, drop in the inputs, write a ~6-line `test.yaml`, run
`pytest --update-benchmarks -k <name>`, check the diff, commit. There is no longer a second
or third place to register the test: both pytest and CMake discover tests by looking for
`test.yaml`.

### Regenerating reference files is a supported operation

`--update-benchmarks` re-runs tests and overwrites their references. It refuses to write a
reference from a run that crashed or that produced nothing parseable, and it warns if you try
to generate references from an MPI run rather than a deterministic serial one.

### Better failure reporting

A failed comparison now lists every value that is out of tolerance (the quantity, its
position, both values and both errors) and prints the path to the work directory so you can
go and look at the actual output.

### One real check added

If a parser extracts nothing from an output file, the test now fails. testcode treated
"no data extracted" as a pass, which meant a test could keep passing forever after an output
format change stopped the parser matching. All 97 current references were checked against
this and none is affected.

### Wannier90 now exits with a nonzero status on a fatal error

This is a change to the programs themselves, not just the tests. A fatal error ended in a
bare Fortran `stop`, which returns 0: so a failed run looked identical to a successful one
to any script, Makefile or CI step. Error paths now return 1. Affects `wannier90.x`,
`postw90.x`, `w90chk2chk.x` and `w90spn2spn.x`. The postprocessing-setup (`-pp`) success path
still returns 0.

Related: when running under MPI, an error code raised on a non-root rank could leave that
rank reporting success while rank 0 reported failure. Library callers saw an inconsistent
return code. All ranks now report the same one.

**This is user-visible**: scripts that assumed Wannier90 always returns 0 will start seeing
failures reported. Flagged in the CHANGELOG under its own heading.

### Build system and CI

- CMake discovers tests by glob instead of a hand-maintained list of 97 names. `ctest` keeps
  one entry per test, and `ctest -L postw90` (and the other categories) works.
- Configuring no longer copies the test data (~240 MB) into the build directory.
- If pytest or PyYAML are missing, CMake warns and skips registering the tests rather than
  failing, so a plain build of the binaries still works.
- CI workflows call pytest, install the Python requirements, and collect the right files as
  artefacts when a run fails.
- Minimum Python raised from 3.8 (end of life since 2024) to 3.10.

### Documentation

`test-suite/README.md` is rewritten around the new workflow. `CONTRIBUTING.md` is updated
(and no longer refers to Travis-CI). `CHANGELOG.md` gains an entry, including the
acknowledgement of J. Spencer's testcode, which served this project well for years.

## Results

| Check                         | New                      | Old (baseline) |
|-------------------------------|--------------------------|----------------|
| Serial run                    | 96 passed, 1 skipped     | 96 passed      |
| MPI run (`--nprocs=2`)        | 97 passed                | 97 passed      |
| `ctest`                       | 97 passed, 0 failed      | —              |
| Parallel (`pytest -n auto`)   | same results, ~5x faster | —              |
| Unit tests for the new code   | 93 passed                | —              |

The single serial skip is `partestw90_mpierr`, which needs at least 2 MPI ranks and now says
so itself instead of being commented out. The old suite achieved the same by excluding it
from the default category.

Every test the old suite ran, the new suite runs and passes. Three tests that were commented
out in `tests/CMakeLists.txt` (with reasons naming old CI toolchains) have been re-enabled
after repeated runs showed no failures; if they turn out to be flaky elsewhere, disabling one
again is a single line in its `test.yaml`.

## Notes and possible follow-ups

- Regenerating all references on an unmodified build produces large diffs. Most of it is
  cosmetic (banner text, timings, dates that no parser reads), but some references are old
  enough to predate several releases and their numbers have drifted, still within tolerance,
  as reported in some recent issues. This will be done as a separate PR to ensure that
  old tests were migrated unchanged.
- `pytest -n auto` was verified to give identical results over repeated runs, but is left off
  in CI for this first PR. Enabling it is a one-line change.
- Using TOML instead of YAML would drop the PyYAML dependency entirely, but needs Python 3.11
  as a floor rather than 3.10. Not done here.
- The parsers were moved and had their Python 2 leftovers removed, but their logic is
  untouched. One genuine bug was fixed in an error message path in `parse_boltzwann.py`.
- `test-suite/library-mode-test*/` and `test-suite/checkpoints/` were deliberately left alone.


JeromeCCP9 (approved):
- fixes error exit status codes
- replaces test system

PR #697, opened 2026-09-09
Build and publish the developer documentation (FORD) automatically

### LINKED ISSUE #4: Code Documentation
From the discussion a proposal is to keep the user guide up to date - as a list of input parameters, functionality, specification of input and output files.

It was suggested to use FORD to document modules and subroutines (useful for coders)
https://pypi.python.org/pypi/FORD

giovannipizzi:
The FORD developer documentation was built and uploaded by hand, and the
published copy had not been updated since 2020. It is now built by CI on every
pull request, and published on every push to develop at
https://wannier-developers.github.io/wannier90/

**Note:** on every PR, the graph (with graphviz) is not compiled, only the docs:
in this way, we catch any syntax error and report as a CI error, but we save a
lot of time: on a macOS laptop, building without a graph took 30 seconds, with graph
5 minutes and 30 seconds (11x slower).
On a merge on the main `develop` branch, the graph is instead built, so that
the published documentation has it.

Two comments in the source confused FORD. One stopped it reading
src/types.F90, so the whole w90_types module was missing from the
documentation; the other lost nine lines of text in src/utility.F90. Both are
fixed. FORD reports success even when it drops a file like this, so the build
now fails if anything was dropped.

README.md and CONTRIBUTING.md link to the new address, and the autodoc README
says how to build the documentation locally.

**Note:** After merging, one setting has to be changed by hand: Settings -> Pages ->
Source = "GitHub Actions". The old copy on wannier.org can then be removed.

Fixes #4

### DIFF
--- a/.github/workflows/autodoc.yaml
+++ b/.github/workflows/autodoc.yaml
@@ -1,0 +1,46 @@
+name: 📗 autodoc
+
+# Publishes the FORD developer documentation to GitHub Pages on every push to
+# develop, and on demand.
+#
+# Pull requests do not deploy: they are built as a regression check by the
+# 📗 autodoc job in ci.yaml, which calls the same reusable build workflow.
+#
+# One-time manual setup is required before the deploy job can succeed: repo
+# Settings -> Pages -> Build and deployment -> Source = "GitHub Actions", and
+# the github-pages environment must allow deployment from develop.
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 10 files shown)


JeromeCCP9:
Is it possible to get pre-commit to flag up instances of the problematic comment string?

JeromeCCP9 (approved):

PR #698, opened 2026-09-09
Fix undefined output value in the parameter report

giovannipizzi:
The "System extended in" line in the parameter report was printed without the
variable behind it ever being set, unless the input file specified a
one-dimensional axis. The line is printed whenever transport is enabled or the
output verbosity is above 2, so runs that did not set the axis wrote out
whatever happened to be in memory at the time.

This produced a stray non-text character in the .wout file, varying from run to
run and from build to build, and made testw90_na_chain_gamma fail
intermittently, since the test reads the output file as text (this showed at
least twice in the Fedora/gcc test - pinging @JeromeCCP9 who saw this).

The variable is now set to "none" by default, which is the value the report was
always meant to show. The width of the field is unchanged, so all existing
output is byte-identical.

The pull request also adds a CI step that saves the test working directory when
a job fails, so output files from an intermittent failure can be inspected
afterwards. It also corrects the description of the `na_chain_gamma` test, which named
the wrong material.

### DIFF
--- a/.github/workflows/step_test.yaml
+++ b/.github/workflows/step_test.yaml
@@ -80,3 +80,12 @@ jobs:
         with:
           workflowPreset: "${{ matrix.toolchain }}-ci"
         continue-on-error: ${{ matrix.experimental && inputs.mask-experimental}}
+      - name: Upload test work directory on failure
+        if: failure()
+        uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


jryates (approved):

JeromeCCP9:
nice catch!

PR #706, opened 2026-09-10
Clarify ordering of projections specified in input

### LINKED ISSUE #463: Order of projections in nnkp file
I noticed that if I specify projections in one line for one atom, the order is not strictly preserved.

Say, if I put  `Fe:s;p`  or `Fe:p;s` - then I get the same : the s orbitals come in the nnkpt file first, and then the p's .Looks like the code orders projections by the number l. Accordingly, the hybrids come first (because the are associated with negative l)

Only if I give two lines, like 
```
Fe:p
Fe:s
```
then I get p's first, s last

So, my questions are 
1. is this intended (if anyone remembers that)
2. is this behavior documented (I did not hind such in the projections section of the user guide)

The order of projections is important in some cases, e.g. in the [symmetrization](https://docs.wannier-berri.org/en/master/docs/system.html#wannierberri.system.System.symmetrize) of the Wannier Hamiltonians.

VVitale:
Clarify that if various angular momentum projections are specified in a single line, their order is not preserved.

Fixes #463


giovannipizzi (approved):
Thanks!

PR #713, opened 2026-09-10
Warning to signal conv_tol is ignored if conv_window is not set

### LINKED ISSUE #357: unexpected default behaviour: tolerance for wannerisation ignored
Hello Wannier Developers,

we noticed that the default behaviour currently seems to be to ignore the convergence threshold for wannierisation (the input variable **conv_tol**), which is a little surprising because it causes the wannierisation to perform the maximum number of iterations.

This is because **conv_window** (the number of times the wannierisation must meet convergence criteria before being accepted as converged) is set by default to -1 in param_read().

I see that the default value for **conv_window** is used unless **conv_noise_amp** is set to some positive value or **conv_window** is specified explicitly.  Various of the test cases do indeed explicitly specify both **conv_tol** and **conv_window**.

Isn't it a little confusing that the default behaviour is to ignore the tolerance and do a fixed number of iterations?  Wouldn't it be better to set **conv_window** be set to a default of, say, 3 and then the behaviour is simpler?

If this is a known peculiarity/feature/issue, then of course please do close this issue directly.

Yours, with many thanks,
Jerome.

VVitale:
By default conv_window is -1, which disables the convergence check entirely -- so conv_tol is silently ignored
and wannierisation always runs for the full num_iter iterations. If one wants the minimisation to stop early
once the spread change is below conv_tol for several consecutive iterations, set conv_window explicitly (e.g.
conv_window = 3).
A warning is printed to output if conv_tol is set, but conv_window is not (if iprint > 0)

This partially fixes #357

### DIFF
--- a/src/wannier90_readwrite.F90
+++ b/src/wannier90_readwrite.F90
@@ -280,7 +280,7 @@ contains
       if (allocated(error)) return
 
       call w90_wannier90_readwrite_read_wannierise(settings, wann_control, num_wann, &
-                                                   stdout, error, comm)
+                                                   stdout, print_output%iprint, error, comm)
       if (allocated(error)) return
 
       call w90_readwrite_read_gamma_only(settings, gamma_only, num_kpts, error, comm)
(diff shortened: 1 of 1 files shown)


giovannipizzi:
It would be good to add the same comment also in the docs

JeromeCCP9 (approved):

PR #725, opened 2026-09-11
Replaces a wrong figure in tutorial solutions with the correct one, trying to keep the original layout

### LINKED ISSUE #343: Fig. 14f in the Solution Booklet might not correct
I reproduced Fig. 14a–e in the Solution Booklet according to the procedure given in the Tutorial. However, I cannot reproduce the result of Fig. 14f.
After further investigation, I find out that in the Solution Booklet, the Wannier90 bandstructure in Fig. 14e is the same as Fig. 14f. Therefore, I think Fig. 14f might not correct. I attached my result for Fig. 14f in here, for which I used the disentanglement parameters as shown below
`dis_win_max       = 45.0`
`dis_num_iter      =  60`
`dis_mix_ratio     = 1.0d0`

Can anybody check this issue for me? Thank you.
Fig14f

VVitale:
Fig. 14(f) in the original solution booklet was wrong. This commit replace the original figure with the a new correct one in the online tutorial solutions. I have tried to keep the layout as close as possible to the original figure. 

This fix #343


giovannipizzi:
Thanks! I'm approving as it's better to have a correct figure than a wrong one, but could you @VVitale also do this?
- regenerate all figures (so even if the style is different, it's OK as all 6 have the same style)
- add the data of the DFT band structure somewhere in the solution folder, so it's easy to regenerate in the future (as a note for the future, I think the DFT calculation is the same that we can find in exercise 6).

giovannipizzi (approved):
See earlier comment, though

giovannipizzi (approved):
See earlier comment, though

giovannipizzi (approved):
See earlier comment, though

PR #733, opened 2026-09-18
Add w90_is_mpi_build() to the library API

jaemolihm:
## Problem

A caller of `libwannier90` cannot discover whether the library it linked was built with MPI.
`w90_set_comm` is not guarded by `W90_MPI`, so a serial build accepts a communicator and then
ignores it; `valid_communicator` returns `.true.` unconditionally in a serial build (correctly — a
serial build needs no communicator), so the two cases are indistinguishable.

An MPI caller on more than one rank against a serial library therefore runs the whole
Wannierisation on every rank, using only the data that rank holds, `distk` ignored, with no error.
The library cannot detect this itself: with no MPI compiled in it cannot size the handle it was
given, so the check has to be available to the caller. Encountered in pw2wannier90 library mode.

## Change

```fortran
logical function w90_is_mpi_build()
```
```c
bool w90_is_mpi_build(void);
```

- No library object argument: the answer does not depend on one, and the check is most useful
  before any exists. Precedent: `w90_get_fortran_stdout`/`_stderr`.
- No `ierr`: it cannot fail.
- A function, not a subroutine: the rest of `w90_library` is subroutines because they must return
  `ierr`.
- `src/comms.F90` is untouched. In `wannier90.h` the declaration is outside the trailing `W90_MPI`
  guard, which is evaluated when the *calling* code compiles.

## Testing

gcc/gfortran 14.2 + OpenMPI 4.1.8: **102/102 serial**, **105/105 MPI** ctest, including all
`library-mode-test*`. `fprettify --indent 2` clean. Assertions added to the two library tests.

## Open question

Function vs subroutine — happy to switch to `subroutine w90_is_mpi_build(mpi_build)` if strict
shape uniformity is preferred; nothing else changes.


JeromeCCP9 (approved):
Nice!

